PR #1, opened 2016-12-13
files for opam installation and toplevel

gasche:
To run a toplevel, first install `utop` (`opam install utop`) and then

```
make runtop
```

will run the toplevel with the Sourir libraries (listed in sourir.mllib) loaded and ready to use.


o-:
lgtm
(what do you think about running the tests in ocamlinit?)

PR #2, opened 2016-12-13
support scope annotations

gasche:
{ x, y, z } <instruction>
    
requires that exactly `{ x, y, z }` be available at this point in the
program. All other variables will be dropped from the environment.
    
      { x, y, ... } <instruction>
    
requires that at least those variables be available.

### DIFF
--- a/tests.ml
+++ b/tests.ml
@@ -12,17 +12,26 @@ let has_var x v =
 
 let (&&&) p1 p2 conf = (p1 conf) && (p2 conf)
 
+let drop_annots = Array.map snd
+
 let run prog pred () =
-  let final_conf = run_forever prog in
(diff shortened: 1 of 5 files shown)


o-:
can you add some testcases?

tests.ml
```diff
@@ -151,10 +160,40 @@ let test_broken_scope_3 =
     print x;
   |]
```
gasche, line 162: @o- see testcases below :-)

o-: @gasche oh I see, I did not realize that the assembler also understands the annotations now. But the parser/lexer is essentially untested, right?
but lets merge this since the parser/lexer is already in anyways.

PR #3, opened 2016-12-13
Run after parse

gasche:
Builds on top of #2 (but the overlap between the two is small).


o-:
If you already want to support the annotated version, can we have tests for that?

o-:
lgtm

PR #4, opened 2016-12-13
make disassembler and parser match up and add tests

o-:


disass.ml
```diff
@@ -0,0 +1,38 @@
+open Instr
+
+let disassemble prog =
```
gasche, line 3: My recommendation would be to print the string to a buffer ([Buffer](http://caml.inria.fr/pub/docs/manual-ocaml/libref/Buffer.html) module documentation), which allows you to use an imperative "print this then that" style and is also more efficient than repeated string concatenations (not that it's really an issue here). You can use `Buffer.add_{char,string}` directly, but `let print = Printf.bprintf buf in .... print "%d" i .. print "%B" b` is probably more convenient (see [Printf](http://caml.inria.fr/pub/docs/manual-ocaml/libref/Printf.html) documentation).

gasche:
This looks fine to me. Do you want to redo the printer in buffer-filling style, or should we merge now and leave as an incremental improvement for later?

disass.ml
o-: Ok, I'll do it together with the annotations. working on that right now

PR #5, opened 2016-12-13
extend disassembler to support annotations

o-:
Based on #4 
This also gets rid of string concatenation


disass.ml
```diff
@@ -0,0 +1,45 @@
...
+        begin match instr with
+        | Decl_const (var, exp)           -> pr buf " const %s = " var; dump_expr exp
+        | Decl_mut (var, exp)             -> pr buf " mut %s = " var; dump_expr exp
+        | Assign (var, exp)               -> pr buf " %s <- " var; dump_expr exp
+        | Branch (exp, l1, l2)            -> pr buf " branch "; dump_expr exp; pr buf " %s %s" l1 l2
+        | Label label                     -> pr buf "%s:" label
+        | Goto label                      -> pr buf " goto %s" label
+        | Print exp                       -> pr buf " print "; dump_expr exp
+        | Invalidate (exp, l, vars)       -> pr buf " invalidate "; dump_expr exp;
+                                             pr buf " %s (%s)" l (String.concat ", " vars)
```
gasche, line 37: the parser doesn't use commas there (just space-separated variables) but I can change this. I agree it's more consistent with scope annotations. The reason I want commas in scope annotations is that we want to be able to extend that to const/mut qualifiers, and `const x, const y, mut z` (or the compact form `const x, y, mut z`) works but (`const x const y mut z`) looks horrible.

disass.ml
```diff
@@ -0,0 +1,45 @@
...
+        | Stop                            -> pr buf " stop"
+        | Comment str                     -> pr buf " %s" str
```
gasche, line 39: Do we keep the `#` in the comment string? We probably should not. Again that sounds like a parser change, but it would also require a change here.

tests.ml
```diff
@@ -245,7 +266,22 @@ let suite =
...
+   "parser5">:: test_parse_disass ("# asdfasdf\n");
+   "parser6">:: test_parse_disass ("branch (x == y) as fd\n");
+   "parser7">:: test_parse_disass ("const x = (y + x)\n x <- (x == y)\n# asdfasdf\nbranch (x == y) as fd\n");
+   "parser8">:: test_parse_disass_file "test.sou";
+   "disass1">:: test_disass_parse (test_sum 10);
+   "disass2">:: test_disass_parse (test_add 1 0);
+   "disass_scope1">:: test_disass_parse test_broken_scope_4;
+   "disass_scope2">:: test_disass_parse test_broken_scope_4_fixed;
+   "disass_scope3">:: test_disass_parse test_broken_scope_5;
+   "parser_scope1">:: test_parse_disass "{a, b} print x\n{a,x,...} #asdf\n";
```
gasche, line 283: I guess we need an `invalidate` test as well. Can we make sure we test all instructions?

(I'll try to think of adding code-coverage instrumentation in the future.)

gasche:
I made a couple comments that we can fix now or after we merge.

disass.ml
o-: oh i see, I like commas

tests.ml
o-: yeah, I refrained from this since we still don't know how exactly the invalidate should behave and be used. As soon as we have the first usecases this should stabilize naturally i guess

PR #6, opened 2016-12-13
Parser printer changes

gasche:
Mostly small fixes for minor issues in #5.


disasm.ml
```diff
@@ -34,9 +34,9 @@ let disassemble (prog : (Scope.scope_annotation option * Instr.instruction) arra
         | Goto label                      -> pr buf " goto %s" label
         | Print exp                       -> pr buf " print "; dump_expr exp
         | Invalidate (exp, l, vars)       -> pr buf " invalidate "; dump_expr exp;
-                                             pr buf " %s (%s)" l (String.concat ", " vars)
+                                             pr buf " %s [%s]" l (String.concat ", " vars)
         | Stop                            -> pr buf " stop"
-        | Comment str                     -> pr buf " %s" str
+        | Comment str                     -> pr buf "#%s" str
```
o-: shouldn't the comment symbol be aligned with the code? the disassembler prints all instructions except labels with a space at the beginning of the line to be able to quickly spot control flow. Imho the comment should belong with the normal code indentation.

o- (approved):

disasm.ml
gasche: That's a good point, I'll change.

PR #7, opened 2016-12-13
read instruction / input tape

gasche:


gasche:
There is a lot of redundancy in the various `string -> stuff -> string` conversions we have, maybe it would make sense to have an `IO` module that depends on `Instr` and has these, on which both `Disasm` and this code (in `Eval`) would depend.

tests.ml
```diff
@@ -235,31 +249,41 @@ let test_disasm_parse prog = function() ->
 let suite =
   let open Assembler in
   "suite">:::
-  ["mut">:: run_checked test_mut
+  ["mut">:: run_checked test_mut no_input
```
o-, line 252: maybe a bit burdensome to have to write no_input in every testcase. But then again we should probably have some way of more programatically injecting tests anyways

o- (approved):
lgtm
regarding the redundancy you mention: By sharing code (it's not that much) we might introduce more coupling, not?
But you are right that we have essentially the same matches in several places. If only we had objects :) (ie. we could: instruction.print_to buf).
Another approach would be template/metaprog. whats the default approach in ocaml there? something like:
match instr with
%foreach Instr | Instr -> "%Instr"

gasche:
I updated the code with a separate `IO` module and the litteral/value string functions factorized in `Instr`. What do you think?

instr.ml
```diff
@@ -54,6 +54,22 @@ type binding =
...
+  | Bool b -> string_of_bool b
+  | Int n -> string_of_int n
+
+let litteral_of_string : string -> litteral = function
+  | "nil" -> Nil
+  | "true" -> Bool true
+  | "false" -> Bool false
+  | n ->
+    try Int (int_of_string n) with _ ->
+      Printf.kprintf invalid_arg "litteral_of_string %S" n
```
gasche, line 68: Explanations: `%S` prints a quoted quotes-and-newline-escaped string (it's good when you want to present unknown string values which may have whitespace etc.), `invalid_arg` is a standard function that just raises the `Invalid_argument exception`, and `kprintf` applies its first argument to the result of formatting the other arguments.

o-:
oh you only wanted those bits a pieces going into IO at the moment. lgtm

PR #8, opened 2016-12-14
factor out analysis algorithm from scope inference

o-:
This introduces a generic forward analysis function

The reason for this patch is to make it easier to write other analysis'.

As a nice side-effect it makes it even more apparent that scope inference closely follows abstract interpretation

### DIFF
--- a/analysis.ml
+++ b/analysis.ml
@@ -1,10 +1,10 @@
 open Instr
 
-let successors prog_at program pc =
+let successors program pc =
   let pc' = pc + 1 in
   let next = if pc' = Array.length program then [] else [pc'] in
-  let resolve = Instr.generic_resolve prog_at program in
-  match prog_at program pc with
+  let resolve = Instr.resolve program in
+  match program.(pc) with
   | Decl_const _
   | Decl_mut _
   | Assign _
@@ -30,14 +30,12 @@ module VarSet = Set.Make(Variable)
 
 let forward_analysis (init_state : 'a)
                      (merge : 'a -> 'a -> 'a option)
-                     (update : 'p -> 'a -> 'a)
-                     (program : 'p array)
-                     (prog_at : 'p array -> int -> Instr.instruction)
+                     (update : int -> 'a -> 'a)
+                     (program : program)
                      : 'a option array =
   let program_state = Array.map (fun _ -> ref None) program in
   let rec work = function
     | (in_state, pc) :: rest ->
-        let instr = program.(pc) in
         let cell = program_state.(pc) in
         let merged =
           match !cell with
@@ -47,8 +45,8 @@ let forward_analysis (init_state : 'a)
         | None -> work rest
         | Some merged ->
             cell := Some merged;
-            let updated = update instr merged in
-            let succs = successors prog_at program pc in
+            let updated = update pc merged in
+            let succs = successors program pc in
             let new_work = List.map (fun pc -> (updated, pc)) succs in
             work (new_work @ rest)
         end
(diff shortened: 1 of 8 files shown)


analysis.ml
```diff
@@ -0,0 +1,58 @@
...
+            cell := Some merged;
+            let updated = update instr merged in
+            let succs = successors prog_at program pc in
+            let new_work = List.map (fun pc -> (updated, pc)) succs in
+            work (new_work @ rest)
+        end
+    | [] -> ()
+  in
+  work [(init_state, 0)];
+  Array.map (fun r -> !r) program_state
```
gasche: shortcut of the day: `(fun r -> !r)` is `(! )`, just like `(fun a b -> a + b)` is `( + )` (the spaces are optional except for infix multiplication, `( * )`, to avoid a clash with the comment syntax `(* ... *)`).

analysis.ml
```diff
@@ -0,0 +1,58 @@
...
+        in begin match merged with
+        | None -> work rest
+        | Some merged ->
+            cell := Some merged;
+            let updated = update instr merged in
+            let succs = successors prog_at program pc in
+            let new_work = List.map (fun pc -> (updated, pc)) succs in
+            work (new_work @ rest)
+        end
+    | [] -> ()
```
gasche: Short cases should be first.

analysis.ml
```diff
@@ -0,0 +1,58 @@
...
+  | Branch (_e, l1, l2) -> [resolve l1; resolve l2]
+  | Stop -> []
+
+module VarSet = Set.Make(Variable)
+
+(* Perform forward analysis on some code
+ *
+ * init_state : Initial input state to instruction 0
+ * merge      : current state -> input state -> merge state if changed
+ * update     : abstract instruction -> input state -> output state
```
gasche, line 24: We could have a record type `type ('a, 'p) domain = ...` for the `init_state`, `merge` and `update` instructions as they make sense to gether -- and that makes sure that producer and consumer uses consistent naming for these functions, that is the record field names.

analysis.ml
```diff
@@ -0,0 +1,58 @@
...
+ * prog_at    : program -> index -> instruction at index
+ *
+ * Returns an array of states for every instruction of the program.
+ * Bottom is represented as None *)
+
+let forward_analysis (init_state : 'a)
+                     (merge : 'a -> 'a -> 'a option)
+                     (update : 'p -> 'a -> 'a)
+                     (program : 'p array)
+                     (prog_at : 'p array -> int -> Instr.instruction)
```
gasche: This feels a bit over-generalized to me. Could we instead have `(update : int -> instr -> 'a -> 'a)`, and then `(program : instr array)` (or even `(int -> instr)` if we prefer) and no `prog_at` function?

gasche (approved):
I think the idea is really neat, but you may have over-generalized the implementation a bit. Could you try to see whether we can remove the `'p` parameter completely? The idea would be to have the `pc` available in `update` and, in the scope-inferrence instance, just lookup a separate `annotations` array from there.

o-:
ok, how about I try to unify the program representations first!

gasche:
You can do it first but it shouldn't be necessary. We want an interface that is generic enough to support different sorts of side-program metadata (although one possibility is to carry separate arrays for metadata : `scope array * instr array` for example instead of `(scope * instr) array`, which is one way to unify), but the question is whether we can have it with a minimal amount of complexity in the generic interface.

analysis.ml
o-: I don't understand this point

o-:
ptal,
I think its a lot easier to deal with the situation if we keep the program separate from the annotations

gasche:
@o- if you are building on top of this branch ( #9 ) we should probably merge it now. Would you solve the conflicts and merge?

o-:
with pleasure

PR #9, opened 2016-12-14
prune false branches transformation

o-:
Based on #8 

Implements a transformation which prunes false branches for invalidate

### DIFF
--- a/transform.ml
+++ b/transform.ml
@@ -4,8 +4,9 @@ let remove_empty_jmp prog =
   let rec remove_empty_jmp pc =
     let pc' = pc + 1 in
     if pc' = Array.length prog then [prog.(pc)] else
+      let pred = Analysis.predecessors prog in
       match (prog.(pc), prog.(pc')) with
-      | (Goto l1, Label l2) when l1 = l2 ->
+      | (Goto l1, Label l2) when l1 = l2 && List.length (pred pc') = 1 ->
           remove_empty_jmp (pc+2)
       | (_, _) ->
           prog.(pc) :: remove_empty_jmp (pc')
--- a/transform.ml
+++ b/transform.ml
@@ -1,0 +1,89 @@
+open Instr
+
+let remove_empty_jmp prog =
+  let rec remove_empty_jmp pc =
+    let pc' = pc + 1 in
+    if pc' = Array.length prog then [prog.(pc)] else
+      match (prog.(pc), prog.(pc')) with
+      | (Goto l1, Label l2) when l1 = l2 ->
+          remove_empty_jmp (pc+2)
+      | (_, _) ->
+          prog.(pc) :: remove_empty_jmp (pc')
+  in
+  let cleaned = remove_empty_jmp 0 in
+  Array.of_list cleaned
+
+let remove_dead_code prog entry=
+  let dead_code =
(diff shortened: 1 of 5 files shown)


transform.ml
```diff
@@ -0,0 +1,89 @@
+open Instr
+
+let remove_empty_jmp prog =
+  let rec remove_empty_jmp pc =
+    let pc' = pc + 1 in
+    if pc' = Array.length prog then [prog.(pc)] else
+      match (prog.(pc), prog.(pc')) with
+      | (Goto l1, Label l2) when l1 = l2 ->
```
gasche: is that correct if other instructions can jump to `l2`? I'd think you need to keep `Label l2` in the output unless you can check that it has a single predecessor.

transform.ml
```diff
@@ -0,0 +1,89 @@
...
+  in
+  let cleaned = remove_empty_jmp 0 in
+  Array.of_list cleaned
+
+let remove_dead_code prog entry=
+  let dead_code =
+    let init_state = ((), entry) in
+    let merge _ _ = None in
+    let update _ _ = () in
+    Analysis.forward_analysis init_state merge update prog
```
gasche, line 22: cute!

transform.ml
```diff
@@ -0,0 +1,89 @@
...
+  in
+  let rec remove_dead_code pc =
+    if pc = Array.length prog then [] else
+      match dead_code.(pc) with
+      | None -> remove_dead_code (pc+1)
+      | Some _ -> prog.(pc) :: remove_dead_code (pc+1)
+  in
+  Array.of_list (remove_dead_code 0)
+
+(* TODO: allow pruning of the true branch, but we need not expression for that *)
```
gasche, line 32: Adding "not" seems tricky design-wise: do we allow (not expression), but then do we have complex expressions in general?
One thing we can do is to have `not variable`, and then make sure that for any boolean primop, we have its negation also as a primop: we have `==`, we can add `!=`. Then we can implement `not` as an expression-rewriting function (`true <-> false`, `x -> not x`, and `x == y <-> x != y`).

transform.ml
```diff
@@ -0,0 +1,89 @@
...
+  (* Convert "branch e l1 l2" to "invalidate e l1; goto l2" *)
+  let rec kill_branch pc =
+    if pc = Array.length prog then [Stop] else
+    match scope.(pc) with
+    | Scope.Dead -> assert(false)
+    | Scope.Scope scope ->
+        begin match prog.(pc) with
+        | Branch (exp, l1, l2) ->
+            let vars = Scope.VarSet.elements scope in
+            Invalidate (exp, "%deopt_" ^ l2, vars) ::
```
gasche: `let depot_label_of l = "%depot_" ^ l` could be an auxiliary function -- hard string constants that occur several times are kind of a code smell.

transform.ml
o-: good catch

transform.ml
o-: I actually already added the != in #10 and also thought about the expression rewriting. but it seems messy, since we'll have to support every expression which evaluates to bool and it will not be clear where the code comes from. Maybe we should have invalidate_on_true and invaliate_on_false?

PR #11, opened 2016-12-15
Optimize find_origins

gasche:
The first commit keeps the two implementations in parallel and check that they return the same result.


gasche:
(I had at first forgotten to update the tests. Maybe we should setup continuous integration testing on pull requests.)

o-:
maybe we should do something similar with resolve?

PR #12, opened 2016-12-15
optimize "resolve" with a program-global table

gasche:
> maybe we should do something similar with resolve?

Maybe, indeed.

I kept a `resolve` function that only resolves one label, as it is
likely to be faster than computing the program-global table for just
one resolution (there is such a use-case in the branch
removal analysis).


o-:
considering we do not care about performance, I am not sure if it was such a great idea, sorry.
Imho #11 was still justifiable since at least it also unified the implementation definition of next and previous instructions, preventing them from getting out of sync. But this one just complicates the api for some performance in some cases...
That being said its a small enough that I will not object it, if you'd like to merge it.

gasche:
No, I agree that the complexity benefit is probably not worth the cost.

One change in this PR that I would like to merge is moving the definition of `pc` into the `Instr` module. There are places where we use `int` today and `pc` would be more informative, but a dependency on `Eval` is not justified. I will merge just that and drop the rest of the PR.

o-:
sounds good

gasche:
I just pushed the `pc` change in 9eb8c01b66cd8a6a40f267414135c9441d5b26f3

PR #17, opened 2016-12-21
implement more analysis

o-:
* generic backwards analysis
* reaching definitions
* usages

### DIFF
--- a/analysis.ml
+++ b/analysis.ml
@@ -26,12 +26,12 @@ let predecessors program =
   done;
   preds
 
-let do_analysis (successors : pc -> pc list)
-                (init_state : ('a * pc) list)
-                (program : program)
-                (merge : 'a -> 'a -> 'a option)
-                (update : pc -> 'a -> 'a)
-                : 'a option array =
+let dataflow_analysis (next : pc -> pc list)
+                      (init_state : ('a * pc) list)
+                      (program : program)
+                      (merge : 'a -> 'a -> 'a option)
+                      (update : pc -> 'a -> 'a)
+                      : 'a option array =
   let program_state = Array.map (fun _ -> ref None) program in
   let rec work = function
     | [] -> ()
@@ -46,8 +46,8 @@ let do_analysis (successors : pc -> pc list)
         | Some merged ->
             cell := Some merged;
             let updated = update pc merged in
-            let succs = successors pc in
-            let new_work = List.map (fun pc -> (updated, pc)) succs in
+            let cont = next pc in
+            let new_work = List.map (fun pc -> (updated, pc)) cont in
             work (new_work @ rest)
         end
   in
@@ -58,13 +58,14 @@ let exits program =
   let rec exits pc : Pc.t list =
     if Array.length program = pc then []
     else
-      ((if List.length (successors program pc) = 0 then [pc] else [])
-        @ exits (pc + 1)) in
+      let is_exit = successors program pc = [] in
+      if is_exit then pc :: exits (pc + 1) else exits (pc + 1)
+  in
   exits 0
 
 let forward_analysis_from init_pos init_state program =
   let successors pc = successors program pc in
-  do_analysis successors [(init_state, init_pos)] program
+  dataflow_analysis successors [(init_state, init_pos)] program
 
 let forward_analysis init_state program =
   forward_analysis_from 0 init_state program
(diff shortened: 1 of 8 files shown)


o-:
I know you prefer the low-tech code motion. We do not have to use this. I just had fun implementing it...

gasche:
I think there is a good case to be made for the high-tech approach, in particular as it is closer to what domain people do. We should definitely merge this nice work, once I'm through a pass of nitpicking for the principle.

analysis.ml
```diff
@@ -26,24 +26,12 @@ let predecessors program =
...
- *
- * Returns an array of states for every instruction of the program.
- * Bottom is represented as None *)
-
-let forward_analysis (init_state : ('a * pc))
-                     (merge : 'a -> 'a -> 'a option)
-                     (update : pc -> 'a -> 'a)
-                     (program : program)
-                     : 'a option array =
+let do_analysis (successors : pc -> pc list)
```
gasche: I think "successors" is confusing here. What about simply `neighbors` or `next`? Also, `do_analysis` is a meh name, what about `analyze`?

analysis.ml
```diff
@@ -58,10 +46,111 @@ let forward_analysis (init_state : ('a * pc))
...
   in
-  work [init_state];
+  work init_state;
   Array.map (!) program_state
+
+let exits program =
+  let rec exits pc : Pc.t list =
+    if Array.length program = pc then []
+    else
+      ((if List.length (successors program pc) = 0 then [pc] else [])
```
gasche: `successors program pc = []` would be more idiomatic (and also happens to be O(1) even if the list is long.)

analysis.ml
```diff
@@ -58,10 +46,111 @@ let forward_analysis (init_state : ('a * pc))
...
+    | Some x -> Defs.add x (InstrSet.singleton pc) defs in
+  let res = forward_analysis Defs.empty prog merge update in
+  match res.(pc) with
+  | None -> InstrSet.empty
+  | Some res ->
+      let rec union a =
+        match a with
+        | [] -> InstrSet.empty
+        | a :: tail -> InstrSet.union a (union tail)
+      in
```
gasche: `List.fold_left InstrSet.union InstrSet.empty a`

instr.ml
```diff
@@ -82,3 +88,61 @@ let resolve (code : program) (label : string) =
...
+  | Branch (e, _l1, _l2) -> expr_vars e
+  | Label _l | Goto _l -> VarSet.empty
+  | Comment _ -> VarSet.empty
+  | Read x -> VarSet.singleton x
+  | Print e -> expr_vars e
+  | Invalidate (e, _l, xs) ->
+    VarSet.union (VarSet.of_list xs) (expr_vars e)
+  | Stop -> VarSet.empty
+
+let defined_vars = function
```
gasche, line 125: I find the plural here a bit confusing. As I was reading code I assumed that this function returned a set of variables.

analysis.ml
```diff
@@ -58,10 +46,111 @@ let forward_analysis (init_state : ('a * pc))
...
+    | None, Some b -> Some b
+    | Some a, Some b -> Some (InstrSet.union a b) in
+  Defs.merge do_merge
+
+let equal_defs =
+  let do_equal a b = InstrSet.equal a b in
+  Defs.equal do_equal
+
+
+let reaching prog pc =
```
gasche: I may be time to start adding documentation comments to say what those functions are doing. In OCaml it is customary to have the documentation comments just before the function declaration (without a blank line), and of the form `(** ... *)` (note the two starts at the beginning), with mere implementation comments of the form `(* ... *)`.

analysis.ml
```diff
@@ -58,10 +46,111 @@ let forward_analysis (init_state : ('a * pc))
...
+        match a with
+        | [] -> InstrSet.empty
+        | a :: tail -> InstrSet.union a (union tail)
+      in
+      let consumed_vars = consumed_vars instr in
+      let definitions_of var = Defs.find var res in
+      let all_definitions = List.map definitions_of consumed_vars in
+      union all_definitions
+
+let used prog pc =
```
gasche: again a code comment above this declaration would help

analysis.ml
```diff
@@ -58,10 +46,111 @@ let forward_analysis (init_state : ('a * pc))
...
+let reaching prog pc =
+  let instr = prog.(pc) in
+  let merge cur_def in_def =
+    let merged = merge_defs cur_def in_def in
+    if equal_defs cur_def merged then None else Some merged in
+  let update pc defs =
+    let instr = prog.(pc) in
+    match defined_vars instr with
+    | None -> defs
+    | Some x -> Defs.add x (InstrSet.singleton pc) defs in
```
gasche: To the unitiated this code really reads as if you were adding `pc` to the set of definitions of `x` in `defs`. Could you add a comment that you are actually shadowing all previous definitions of `x`, that they will be dropped from the map?

gasche (commented):
I find it a bit frustrating that these functions `reaching` and `used` compute data over the whole program, only to throw it away and just compute the answer for a single `pc`. Could the post-analysis post-treatment be done in an `Array.map` call over each pc position, and then you would use `(used prog).(pc)` instead of `used prog pc`, which is still convenient, but preserves the opportunity for program-level caching?

(Another way to do this is to write the program as `let used prog = ...(* computation *)... fun pc -> result.(pc)`, but I find it a bit too clever for its own good.)

o-:
> (Another way to do this is to write the program as let used prog = ...(* computation *)... fun pc -> result (pc), but I find it a bit too clever for its own good.)

decided to go down that road. It feels very natural and we can communicate the intended usage quite easily through the type signature imho

analysis.ml
o-: ok

analysis.ml
o-: ok

analysis.ml
o-: totally agree, also I now renamed Defs to VariableMap and made it a bit more clear what its intended usage is

analysis.ml
o-: agreed, i added a variable name to communicate this

analysis.ml
o-: thanks! I was fighting the compiler there

analysis.ml
o-: ok

instr.ml
o-: I changed it now to return a list and the code seems to become cleaner. What do you think?
Some people find it confusing to have 0 or 1 sized collections... But I actually often find it more natural than the maybe monad.

gasche: Yes, it's fine: the return type now corresponds to the function name.

> Some people find it confusing to have 0 or 1 sized collections... 

Clearly you are thinking ahead, planning for the introduction of pattern matching to our language.

gasche:
Thanks, the changes are nice, I merged.

Two minor comments:

- in OCaml it is customary to use `function` when there are several cases that follow (`function None -> true | Some foo -> test foo)`), and `fun` when there is a single case (`fun x ->`, `fun () ->` -- or `fun x y () z ->`). Your preference for `function` is like the accent of a foreign speaker: cute (I hope).

- When I asked for comments on the new analyses you introduced, I was thinking of maybe more detailed comments on what the terms mean or of how they are computed (to help follow along the implementation maybe?). What we have now is fine, but obviously the comment as-is are useless, and I apologize for having you write them :p

PR #18, opened 2016-12-21
drive by cleanups

o-:


disasm.ml
```diff
@@ -14,8 +14,8 @@ let disassemble_annotated (prog : Scope.annotated_program) =
     in
     let str_from_vars vars = String.concat ", " (Scope.VarSet.elements vars) in
     begin match annot with
-    | Some (Scope.Exact e)    -> pr buf "{%s} "      (str_from_vars e)
-    | Some (Scope.At_least e) -> pr buf "{%s, ...} " (str_from_vars e)
+    | Some (Scope.Exact e)    -> pr buf "{%s}"      (str_from_vars e)
+    | Some (Scope.At_least e) -> pr buf "{%s, ...}" (str_from_vars e)
```
gasche: not sure about this change; wouldn't you get `{x, y}foo:`, with no space in between, in the label case?

We could, of course, also align the printing of the the instructions (by precomputing the scope part, computing the largest scope annotation in the program, and padding with it), but that sounds overkill.

transform.ml
```diff
@@ -8,11 +8,12 @@ let remove_empty_jmp prog =
       match (prog.(pc), prog.(pc')) with
       | (Goto l1, Label l2) when l1 = l2 && List.length pred.(pc') = 1 ->
           remove_empty_jmp (pc+2)
+      | (Label _, _) when pred.(pc') = [pc-1] ->
+          remove_empty_jmp pc'
```
gasche, line 18: nice :-)

disasm.ml
o-: you are right. this code is ugly. let's not touch it :)

transform.ml
gasche: Looking at this code against I'm mildly confused as to why it is `pred.(pc')` and `pc-1` -- if I understand correctly, `pc-1` is `pc'-2`.

o-: meh, you are right again and it even breaks in the case of
branch ... l1
l1:

transform.ml
```diff
@@ -2,16 +2,17 @@ open Instr
...
   let pred = Analysis.predecessors prog in
+  let succ = Analysis.successors prog in
   let rec remove_empty_jmp pc =
     let pc' = pc + 1 in
     if pc' = Array.length prog then [prog.(pc)] else
       match (prog.(pc), prog.(pc')) with
       | (Goto l1, Label l2) when l1 = l2 && List.length pred.(pc') = 1 ->
           remove_empty_jmp (pc+2)
-      | (Label _, _) when pred.(pc') = [pc-1] ->
+      | (Label _, _) when pred.(pc) = [pc-1] && succ (pc-1) = [pc] ->
```
gasche, line 12: Why do we need to test both conditions? Does the latter work correctly when `pc` is `0`?

o-: 1. we need it because the single predecessor could be a branch:

     branch x l1 l2
    l1:

2. it does work but the reason is not obvious. The first check will be pred.(0) = [-1], therefore not fail but but certainly be false. therefore the second condition (which would fail at succ (-1)) will not be executed...

gasche: I see. Could you add a comment to mention at least the branch case?

gasche:
Ok to go. Could you fix the conflicts and merge?

PR #19, opened 2016-12-21
add a readme

o-:


README.md
```diff
@@ -0,0 +1,35 @@
+# sourir
+
+sourir is better than [RIR](https://github.com/reactorlabs/rir)
```
gasche, line 3: better? smaller? younger? (I'm not sure about faster, stronger)

README.md
```diff
@@ -0,0 +1,35 @@
...
+    opam install ocaml.4.04.0
+    git clone https://github.com/reactorlabs/sourir
+    cd sourir
+    make sourir
+    ./sourir test.sou
+
+## Syntax / Semantics
+
+The language is still in flux and there is no written syntax or semantics. See
+parser.mly or the testcases in tests.ml to get an idea of the syntax. See
```
gasche, line 18: @mhyee looked at the tests for syntax inspiration and he was a bit confused by some of the macros we use there (`eq`, etc.) that are not actual syntax. Maybe we could add a `print-tests` mode where we pretty-print the tested programs, to give more inspirations. Or we may want to merge his "examples" branch ( @mhyee why did you not send a PR for this one?) to have more `.sou` files for people to look at.

gasche (approved):

README.md
mhyee: Yeah, "smaller" might be a more helpful description.

README.md
mhyee: Some of the tests are written with macros, but others directly call the parser.

(I'll open a PR for my examples once I get into the lab and update my branch.)

mhyee:
Do you want to say anything about Merlin? Is `menhir` required?

EDIT: I guess `utop` is also required for `make runtop`.

gasche:
Yes `menhir` is required to build the parser -- so it is required. I'll add a blurb about Merlin later.

o-:
hmm I am reluctant about making this an ocaml tutorial :) Maybe we could have a separate readme for that?

README.md
o-: ok, or we just get rid of the assembler, or is there a good reason to keep it? I already converted some of the examples, since I broke annotations in the parser and didn't bother to fix.

o-: +1 for merging the examples
maybe add a make target test_examples that runs them all, to avoid breaking them

mhyee: I'll add a make target in my PR.

o-:
@gasche is there an ocaml way of specifying the dependencies? It would be nice to have something automated, eg. "opam instal-deps" or "opam install-dev-deps" and it picks the right versions and all... 

or if that does not exist maybe just "make isntall-deps"

(Also I recall having to install ocamlfind too)

gasche:
There are dependencies specified in the `opam` file. So in theory (sorry for the arcane incantations), `opam pin add sourir . --no-action` (this tells the package manager that there is a package `sourir` defined locally) followed by `opam install --deps-only sourir` should work. We could have that in a `install-deps` target, and we need to update the `opam` file with dependencies -- will do.

README.md
gasche: > ok, or we just get rid of the assembler, or is there a good reason to keep it?

It's your baby, I don't tell other parents what to do.

o-:
@gasche ok, I left you some FIXME in the readme

PR #20, opened 2016-12-21
Install deps target

gasche:
from discussion in #19

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,5 +26,9 @@ run: sourir
 clean:
 	ocamlbuild -clean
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -26,5 +26,9 @@ run: sourir
 clean:
 	ocamlbuild -clean
 
+install-deps:
```
o-, line 29: perfect!
I left FIXMEs in the readme.  can you  insert the "make install-deps" there

o- (approved):

PR #22, opened 2016-12-21
Codemotion examples

mhyee:
- Add the codemotion examples
- Move `test.sou` to `examples/sum.sou`
- Update Makefile and README
- New make target for building examples

Note that `no_cm_01.sou` is now `cm_04_before.sou` because we can move the entire branch. I've added a new `no_cm_01.sou` where you cannot move the branch.

The make target simply builds the examples and runs them with `0` as input and throws away stdout. It's basically to test the parser. Is this OK, or should I do something different?


o-:
for now I think just running them is already good enough. We might want to compare the output -- but that's for later.

lgtm

o- (approved):

o-:
just a nitpick, if the branch is not public yet it is nice to use 'git rebase master' instead of 'git merge master'

mhyee:
Thanks, I'll keep that in mind. (It was a bit messy this time because I was transplanting the branch from my copy of the repo to this one, and it was already public.)

o-:
i see. to my own fork I just "git push -f" if I want to rebase. by public I meant when creating the pl

PR #23, opened 2016-12-21
Immediate arguments

gasche:


disasm.ml
```diff
@@ -3,13 +3,16 @@ open Instr
...
-      | Op (Plus, [a; b]) -> pr buf "(%s + %s)" a b
-      | Op (Neq,  [a; b]) -> pr buf "(%s != %s)" a b
-      | Op (Eq,   [a; b]) -> pr buf "(%s == %s)" a b
+    in
+    let dump_expr exp =
+      match exp with
+      | Simple e          -> simple buf e
+      | Op (Plus, [a; b]) -> pr buf "(%a + %a)" simple a simple b
+      | Op (Neq,  [a; b]) -> pr buf "(%a != %a)" simple a simple b
+      | Op (Eq,   [a; b]) -> pr buf "(%a == %a)" simple a simple b
```
gasche, line 15: @o- : `%a` formatter takes two parameter, the first of type `(Buffer.t -> 'a -> unit)` (in the case of `bprintf`) and the second of type `'a`. I'll eventually rewrite the other parts of the disassembler to use this style if it works for you.

mhyee (approved):
Looks good to me.

It might be tedious to make these changes, but there's also the code motion examples that use variables instead of literals.

gasche:
I wrote this branch in the bus when coming to the office, you hadn't sent your PR yet -- otherwise I would have felt obligated to also update the examples as well. Missed opportunity! I'll merge now because "why not", and we can change examples as we feel like it (or not) over time.

disasm.ml
o-: kinda neat

PR #25, opened 2016-12-23
Error messages

gasche:
The way the error messages are obtained is by asking the parser generator for the list of its error states (and the inputs that lead it there, in a sense), and writing a message for each different case by hand. There are 32 error states in the current grammar, so writing 32 messages was a tad tedious, but then I got lost on the green line tonight so I had plenty of time on my hands.

The plumbing parts where I obtain the automaton error state number is admittedly rather arcane, as well as the build-system plumbing to obtain the error state information, compile the hand-written messages to an OCaml function, and update the messages file when the grammar change. But I think the result is relatively neat:

```
File "test.sou", line 3, characters 8-13, parsing error:
>   const limit 10
Parsing an instruction, we parsed "const <var>" so far; the equal sign
"=" is now expected to construct a constant declaration
"const <var> = <expr>".
```

```
File "test.sou", line 6, characters 0-9, parsing error:
> loop_body
Parsing an instruction, we parsed an identifier so far (variable or label). 
- if this is a label declaration, we expect a semicolon: "<label>:"
- if this is an assignment, we expect a left arrow: "<var> <- <expression>"
```

```
File "test.sou", line 9, characters 10-11, parsing error:
>   i <- (i + (1 + 2))
Parsing an expression, we parsed "( <arg> <op>" so far; an argument
(variable or literal value) is now expected to construct an expression
"( <arg> <op> <arg> )".
```

```
File "test.sou", line 9, characters 12-13, parsing error:
>   i <- (i + 1 + 2)
Parsing an expression, we parsed "( <arg> <op> <arg>" so far;
a closing parenthesis ")" is now expected.
```

### DIFF
--- a/parse.ml
+++ b/parse.ml
@@ -1,15 +1,11 @@
-let position {Lexing.pos_fname; pos_lnum; pos_cnum; pos_bol} =
-  let file = pos_fname in
-  let line = pos_lnum in
-  let character = pos_cnum - pos_bol in
-  (file, line, character)
+(** Parsing functions *)
 
-let split_annotations program =
-  let annotations = Array.map fst program in
-  let instructions = Array.map snd program in
-  (instructions, annotations)
+type parse_error =
+  | Lexing of string * Lexing.position
+  | Parsing of message option * Lexing.position * Lexing.position
+and message = string
 
-exception Parse_error of string option * Lexing.position * Lexing.position
+exception Error of parse_error
 
 let parse lexbuf =
   (* see the Menhir manual for the description of
@@ -17,7 +13,7 @@ let parse lexbuf =
   let open MenhirLib.General in
   let module Interp = Parser.MenhirInterpreter in
   let input = Interp.lexer_lexbuf_to_supplier Lexer.token lexbuf in
-  let success prog = split_annotations prog in
+  let success prog = prog in
   let failure error_state =
     let env = match error_state with
       | Interp.HandlingError env -> env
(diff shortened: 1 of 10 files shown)


parse.ml
```diff
@@ -9,9 +9,44 @@ let split_annotations program =
...
+        | Not_found -> None in
+      raise (Parse_error (message, start_pos, end_pos))
+  in
+  Interp.loop_handle success failure input
+    (Parser.Incremental.program lexbuf.Lexing.lex_curr_p)
+
 let parse_string str =
   let lexbuf = Lexing.from_string str in
-  split_annotations (Parser.program Lexer.token lexbuf)
+  parse lexbuf
```
o-, line 36: I know it's actually my leftovers -- but if you don't mind it would be nice to somehow have better errors for parse_string too

o- (commented):
Don't exactly understand the mechanism. But I guess that's ok. I love the result.

parse.ml
gasche: Indeed, I can work on that. `parse_string` already uses the same auxiliary function. The reason why I haven't shared the error-reporting between `parse_file` and `parse_string` is that I first tried that, and that the `exit <error-code>` things break the testsuite (they stop the tests instead of continuing with other tests) while raising an uncaught exception does not. So we need something a bit more subtle, where the error-message-printing is done in both cases, but only one of them aborts. (Or maybe none, and then the users of `parse_file` with also die with an uncaught exception, which is maybe just fine.)

gasche:
@o- I tried to clarify the `Parse` interface in cd1364258878e1903e7fb21907d88b4bc4bfecec, tell me what you think.

P.S.: `.mli` files are interface files that describe the functions exposed by a given module. Besides documentation, they also serve as access control: only the types and values mentioned in the interface are accessible from outside. Large OCaml projects prefer to have `.mli` files for all `.ml` files, but it makes sense to grow them on-demand (when we start feeling that an interface boundary should be better specified) for a young project.

gasche:
I decided to go ahead and merge given that I have another parser change in the pipeline -- support for empty lines.

o-:
oh wow, that's now the de-luxe version :) thanks

PR #26, opened 2016-12-23
fix analysis results for read instruction

o-:


instr.ml
```diff
@@ -132,8 +132,15 @@ let free_vars = function
 let defined_vars = function
   | Decl_const (x, _)
   | Decl_mut (x, _)
-  | Assign (x ,_) -> [x]
-  | _ -> []
```
gasche, line 136: I'm not sure whether I already made a nitpicky comment about these kind of "fragile patterns" in the past, but there we are :p It is possible to enable a warning against them, maybe I'll do that. There may be other warnings that are not set by default and would help.

PR #28, opened 2016-12-23
Parse newlines

gasche:
While working on this PR I noticed that branch pruning currently generates labels of the form `%foo`, which are not accepted by the parser. I understand now why @mhyee tried to write all his labels in his syntax previously.

@o-, @mhyee, should we distinguish labels from identifiers by asking that all labels start with `%`? Or should we allow labels to start with `%`? Or should we change the code to not generate `%`?

### DIFF
--- a/parser.mly
+++ b/parser.mly
@@ -25,7 +25,7 @@ let scope_annotation (mode, xs) =
 %%
 
 program:
-| prog=list(instruction_line) EOF
+| optional_newlines prog=list(instruction_line) EOF
   {
     let annotations, instructions = List.split prog in
     (Array.of_list instructions,
@@ -33,11 +33,13 @@ program:
   }
 
 instruction_line:
-| a=scope_annotation i=instruction NEWLINE { (a, i) }
+| a=scope_annotation i=instruction NEWLINE optional_newlines { (a, i) }
 
 scope_annotation:
 | { None }
-| annot=delimited(LBRACE, scope, RBRACE) { Some (scope_annotation annot) }
+| annot=delimited(LBRACE, scope, RBRACE) optional_newlines { Some (scope_annotation annot) }
+
+optional_newlines: list(NEWLINE) { () }
 
 scope:
 | x=variable COMMA sc=scope { let (mode, xs) = sc in (mode, x::xs) }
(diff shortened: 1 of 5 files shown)


o-:
@gasche the '%' was on purpose. since the user cannot generate such a label we can be sure it does not clash with any other label in the program. (ie. the %label notation is not part of the surface-language but intended only for compiler transformations)

also, I'm not a big fan of syntax level distinguishing different classes of identifiers, but would not be super opposed either...

parser.mly
```diff
@@ -25,15 +25,15 @@ let scope_annotation (mode, xs) =
...
+| list(NEWLINE) prog=list(instruction_line) EOF
   {
     let annotations, instructions = List.split prog in
     (Array.of_list instructions,
      Array.of_list annotations)
   }
 
 instruction_line:
-| a=scope_annotation i=instruction NEWLINE { (a, i) }
+| a=scope_annotation i=instruction nonempty_list(NEWLINE) { (a, i) }
```
o-: would it be unambiguous (make sense) to also allow newlines between scopes and instruction?

gasche: I think that we could even consider scopes as whitespace and ignore them wherever, allowing
`branch x l1 l2 l1: z <- (x + 1) l2:` on a single line -- in particular the fact that primops are parenthesized makes our grammar highly unambiguous. The question is what we want.

One thing that we could do is introduce a semicolon to explicitly conclude instructions, and then allow newlines everywhere (for example `branch (x == x2)\nl1 l2`. Or we could allow to escape newlines to explicitly continue on the next line, à la Python. Or just allow newlines between scopes and instructions, as you suggest.

gasche:
I don't really like the idea that we write code that we cannot disassemble and parse back. I found out about this as I edited a test of `tests.ml` to add newlines just for fun. The `assert_equal string (disassemble code)` test failed, so I turned it into `let assert_equal_code str1 str2 = assert_equal (parse str1) (parse str2)`, and I started getting parsing errors.

One thing that we can easily do is collect the set of labels used in a program. Then we have a function `free_label label` that returns the first of `label`, `label ^ "1"`, `label ^ "2"` etc. that is not already used.

gasche:
I added support for empty lines in-between the scope annotation and the instruction proper. Feel free to suggest other parsing approaches (I think that many an explicit `;` plus ignored-whitespace may in fact be better), it's not hard to implement.

PR #30, opened 2017-01-05
More precise scoping and undefined variables

o-:
This commit introduces undefined variables on the Heap. They can
be declared by "mut x" (ie. a mut declaration without the definition
part). Every heap cell is now | Value v | Undefined.

To deal with the increased possibility for errors the Scope.infer
now keeps track of initialization (ie. first definition) of variables
and raises an exception on use-before-define. The api of infer is
unchanged the additional information is only internal.

Additionally Scope.infer rejects shadowing (ie. declaration of a
previously declared variable).

### DIFF
--- a/analysis.ml
+++ b/analysis.ml
@@ -90,6 +90,7 @@ module InstrSet = Set.Make(Pc)
  * defining a certain variable *)
 module VariableMap = struct
   include Map.Make(Variable)
+  module KeySet = Set.Make(Variable)
 
   (* merge is defined as the union of their equally named sets *)
   let merge =
@@ -100,6 +101,16 @@ module VariableMap = struct
       | None, Some b -> Some b
       | Some a, Some b -> Some (InstrSet.union a b) in
     merge merge_one
+
+  let inter a b =
+    let keys_a = KeySet.of_list (List.map fst (bindings a)) in
+    let keys_b = KeySet.of_list (List.map fst (bindings b)) in
+    let intersection = KeySet.elements (KeySet.inter keys_a keys_b) in
+    let append cur var = add var (InstrSet.union (find var a) (find var b)) cur in
+    List.fold_left append empty intersection
+
+  let singleton var loc =
+      add var (InstrSet.singleton loc) empty
 
   let equal =
     let is_equal a b = InstrSet.equal a b in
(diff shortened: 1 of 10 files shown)


scope.ml
```diff
@@ -14,40 +14,71 @@ type annotated_program = (program * scope_annotation option array)
...
-      | Some (Exact vars) -> VarSet.inter set vars
-      end in
-    let bound = Instr.bound_vars instr in
-    VarSet.union bound constr_set in
-  let res = Analysis.forward_analysis VarSet.empty instructions merge update in
-  let finish pc preset =
+    let update = { declared = Instr.declared_vars instr;
+                   defined = Instr.defined_vars instr } in
+    let shadowed = VarSet.inter cur.declared update.declared in
+    if not (VarSet.is_empty shadowed) then raise (DuplicateVariable shadowed)
```
o-, line 50: If you prefer I can put this line in a different PR

analysis.ml
```diff
@@ -101,6 +102,16 @@ module VariableMap = struct
       | Some a, Some b -> Some (InstrSet.union a b) in
     merge merge_one
 
+  let inter a b =
```
gasche: This could be given a much shorter (and probably more efficient) definition using [`Map.merge`](http://caml.inria.fr/pub/docs/manual-ocaml/libref/Map.S.html#VALmerge), but `merge` above shadows the name `merge`, so we should rename into `union` I guess.

gasche (commented):
The implementation looks fine, and I am ok with forbidding shadowing.

gasche:
Do we need to keep an optional declaration for `mut` variable? We could, in the interest of simplicity, always have `mut` without declaration, and assign right after it when we want one.

You went the elaborate route of adding a notion of "undefined" variable and having some static analysis to prevent reads of undefined variables. A simpler route would have been to just set undeclared mutables to the value `nil`, but now that you put the hard work we should use it.

gasche:
P.S.: If renaming `merge` into `union` is confusing because it maps to the `merge` operation of our abstract domains, then we could rename them into `join`.

o-:
Fixed the VariableMap naming confusion

> Do we need to keep an optional declaration for mut variable? We could, in the interest of simplicity, always have mut without declaration, and assign right after it when we want one.

I would like to keep the instruction "mut x = 3" in the language for symmetry. So I tried to implement it as syntactic sugar. But that makes the implementation worse (ie. its annoying if pc and loc does not correspond anymore). I think we should keep surface language and Instruction module a bijection.

Therefore I suggest to either keep it optional or also remove it from the surface language. What do you prefer?

gasche:
I was thinking of removing it from the surface language indeed, but you say you like keeping the `mut x = 3` instruction for symmetry (with `const x = 3` I suppose?), so let's keep it that way for now. We can always simplify later if it gets annoying.

PR #31, opened 2017-01-10
minor simplification to the abstract analysis code

gasche:
The (array -> references) indirection used to make sense in the first
non-abstract implementations, but now it is compact enough that using
the mutability of arrays directly is simpler and actually more readable.


o-:
Oh nice

PR #32, opened 2017-01-11
liveness analysis

o-:
Refactor used variables analysis a bit. Liveness analysis falls out for free...

### DIFF
--- a/examples/sum.sou
+++ b/examples/sum.sou
@@ -1,7 +1,7 @@
   mut i = 0
   mut sum = 0
-  const limit = 10
-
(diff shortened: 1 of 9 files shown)


examples/sum.sou
```diff
@@ -1,7 +1,7 @@
   mut i = 0
   mut sum = 0
-  const limit = 10
-
+  mut limit = 0
```
gasche, line 3: we can use just `mut limit` now :-)

gasche:
Very nice!

PR #33, opened 2017-01-11
cleanup the branch pruning algorithm

o-:
This should make it more transparent what's going on

### DIFF
--- a/transform.ml
+++ b/transform.ml
@@ -78,10 +78,10 @@ let copy_fresh global_labels prog =
     match todo with
     | [] -> LabelMap.empty
     | l :: tl ->
-      let fresh = next_fresh_label labels l in
-      let labels = LabelSet.add fresh labels in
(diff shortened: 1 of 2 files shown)


transform.ml
```diff
@@ -44,65 +44,107 @@ let remove_dead_code prog entry=
...
+ * returns a copy of the program with all labels fresh and
+ * an updated list of all occurring labels *)
+let copy_fresh global_labels prog =
+  let prog_labels = collect_labels prog in
+  let rec freshened_labels labels todo =
+    match todo with
+    | [] -> LabelMap.empty
+    | l :: tl ->
+      let fresh = next_fresh_label labels l in
+      let labels = LabelSet.add fresh labels in
```
mhyee: For clarity, could you call this `labels'` instead of shadowing the old `labels`?

o-: ok

PR #34, opened 2017-01-11
cfg analysis

o-:
CFG analysis

Basic blocks have a list of the successor BBs. We could also look at the exit instruction to determine the successors on the fly. For efficiency and convenience I decided to precompute it.

Basic block ids are increasing on fwd edges for cheap least dominator discovery.


gasche:
Sorry for the delay. I reviewed the code and it is fine, so I'm merging now.

PR #36, opened 2017-01-17
Codemotion first iteration

o-:
Based on #34 

there might be dragons


o-:
there are many issues I have to resolve first

PR #37, opened 2017-01-19
Codemotion and fixes

o-:
sorry for this crazy big PR! I think it would be hard to entangle it. I can walk you through the changes though...

1. needs cfg. therefore its based on #34 
2. deals with the fact that we might have an already optimized program as input. we do not want to revisit the deopt targets (first of all pruning this code was our initial goal. secondly the invalidate is not symmetric in both branches (ie. the scope in the deopt case is different)) See commit ff00ac9 for the details.
3. fixes the scope of landing pads: this was horribly broken, see a5d5772 for the details
4. finally implements codemotion. for correct codemotion we need to be able to rename variables. since we want to be able to do this in the main program only and not in the deopt code I extended invalidate to support different in- and out- scopes

### DIFF
--- a/.ocamlinit
+++ b/.ocamlinit
@@ -25,8 +25,8 @@ let print_program ppf (program : Instr.program) =
   Format.fprintf ppf "%s" (Disasm.disassemble program)
 ;;
 
-let print_an_program ppf (program : Instr.annotated_program) =
-  Format.fprintf ppf "%s" (Disasm.disassemble_annotated program)
+let print_an_program ppf (program : Instr.program) =
+  Format.fprintf ppf "%s" (Disasm.disassemble program)
 ;;
 
 #install_printer print_set_vars;;
--- a/.ocamlinit
+++ b/.ocamlinit
@@ -1,8 +1,36 @@
-let print_set ppf (set : Instr.VarSet.t) =
+let print_set_vars ppf (set : Instr.VarSet.t) =
   Instr.VarSet.elements set
    |> List.map (Printf.sprintf "%S")
    |> String.concat "; "
    |> Format.fprintf ppf "[%s]"
 ;;
 
-#install_printer print_set;;
+let print_set_vars2 ppf (set : Cfg.BasicBlockSet.t) =
+  let open Cfg in
+  List.map (fun (e : basic_block) -> e.id) (BasicBlockSet.elements set)
+   |> List.map (Printf.sprintf "%d")
+   |> String.concat ";"
+   |> Format.fprintf ppf "[%s]"
+;;
+
+let print_set_vars3 ppf (set : Analysis.InstrSet.t) =
+  let open Analysis in
+  Analysis.InstrSet.elements set
+   |> List.map (Printf.sprintf "%d")
+   |> String.concat ";"
+   |> Format.fprintf ppf "[%s]"
+;;
+
+let print_program ppf (program : Instr.program) =
+  Format.fprintf ppf "%s" (Disasm.disassemble program)
+;;
+
+let print_an_program ppf (program : Instr.annotated_program) =
+  Format.fprintf ppf "%s" (Disasm.disassemble_annotated program)
+;;
+
+#install_printer print_set_vars;;
+#install_printer print_set_vars2;;
+#install_printer print_set_vars3;;
+#install_printer print_program;;
+#install_printer print_an_program;;
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,8 @@ run: sourir
 
 test_examples: sourir
 	for f in examples/*.sou; do yes 0 | ./sourir $$f > /dev/null; done
+	for f in examples/*.sou; do yes 0 | ./sourir --prune $$f > /dev/null; done
+	for f in examples/*.sou; do yes 0 | ./sourir --cm $$f > /dev/null; done
 	for f in examples/*.sou; do yes 0 | ./sourir --cm --prune $$f > /dev/null; done
 
 clean:
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ run: sourir
 
 test_examples: sourir
 	for f in examples/*.sou; do yes 0 | ./sourir $$f > /dev/null; done
+	for f in examples/*.sou; do yes 0 | ./sourir --cm --prune $$f > /dev/null; done
 
 clean:
 	ocamlbuild -clean
--- a/Makefile
+++ b/Makefile
@@ -23,11 +23,19 @@ runtop: lib
 run: sourir
 	./sourir examples/sum.sou
 
+TEMPDIR := $(shell mktemp -d)
+
 test_examples: sourir
-	for f in examples/*.sou; do yes 0 | ./sourir $$f > /dev/null; done
-	for f in examples/*.sou; do yes 0 | ./sourir --prune $$f > /dev/null; done
-	for f in examples/*.sou; do yes 0 | ./sourir --cm $$f > /dev/null; done
-	for f in examples/*.sou; do yes 0 | ./sourir --cm --prune $$f > /dev/null; done
+	mkdir $(TEMPDIR)/examples
+	for f in examples/*.sou; do yes 0 | ./sourir --quiet $$f              > $(TEMPDIR)/$$f.out; done
+	for f in examples/*.sou; do yes 0 | ./sourir --quiet --prune $$f      > $(TEMPDIR)/$$f.opt.out && diff $(TEMPDIR)/$$f.out $(TEMPDIR)/$$f.opt.out; done
+	for f in examples/*.sou; do yes 0 | ./sourir --quiet --cm $$f         > $(TEMPDIR)/$$f.opt.out && diff $(TEMPDIR)/$$f.out $(TEMPDIR)/$$f.opt.out; done
+	for f in examples/*.sou; do yes 0 | ./sourir --quiet --cm --prune $$f > $(TEMPDIR)/$$f.opt.out && diff $(TEMPDIR)/$$f.out $(TEMPDIR)/$$f.opt.out; done
+	for f in examples/*.sou; do yes 1 | ./sourir --quiet $$f              > $(TEMPDIR)/$$f.out; done
+	for f in examples/*.sou; do yes 1 | ./sourir --quiet --prune $$f      > $(TEMPDIR)/$$f.opt.out && diff $(TEMPDIR)/$$f.out $(TEMPDIR)/$$f.opt.out; done
+	for f in examples/*.sou; do yes 1 | ./sourir --quiet --cm $$f         > $(TEMPDIR)/$$f.opt.out && diff $(TEMPDIR)/$$f.out $(TEMPDIR)/$$f.opt.out; done
+	for f in examples/*.sou; do yes 1 | ./sourir --quiet --cm --prune $$f > $(TEMPDIR)/$$f.opt.out && diff $(TEMPDIR)/$$f.out $(TEMPDIR)/$$f.opt.out; done
+	rm -rf $(TEMPDIR)
 
 clean:
 	ocamlbuild -clean
--- a/analysis.ml
+++ b/analysis.ml
@@ -185,3 +185,4 @@ let used prog : pc -> InstrSet.t =
         let uses_of var = VariableMap.at var res in
         let all_uses = List.map uses_of defined in
         List.fold_left InstrSet.union InstrSet.empty all_uses
+
--- a/analysis.ml
+++ b/analysis.ml
@@ -144,7 +144,7 @@ let reaching prog : pc -> InstrSet.t =
   let update pc defs =
     let instr = prog.(pc) in
     (* add or override defined vars in one go*)
-    let kill = VarSet.elements (defined_vars instr) in
+    let kill = TypedVarSet.vars (defined_vars instr) in
     let loc = InstrSet.singleton pc in
     let replace acc var = VariableMap.add var loc acc in
     List.fold_left replace defs kill
@@ -169,7 +169,7 @@ let liveness_analysis prog =
   let update pc uses =
     let instr = prog.(pc) in
     (* First remove defined vars *)
-    let kill = VarSet.elements (defined_vars instr) in
+    let kill = TypedVarSet.vars (defined_vars instr) in
     let remove acc var = VariableMap.remove var acc in
     let uses = List.fold_left remove uses kill in
     (* Then add used vars *)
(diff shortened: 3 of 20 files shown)


gasche:
I have the impression that the principal problem is the scoping of invalidation instructions and their landing pads. This comes down to discussions we had in the past about the design of invalidation. There are three parts involved:

1. The "invalidate" instruction that is part of the fast-path instruction stream.
2. The "landing pad" that is currently only informally defined as an on-the-side instruction stream that is "uniquely owned" by the invalidate instruction, in the sense that it is its only predecessor. Its purpose is to help with eventual adjustments required before jumping to...
3. The "unoptimized code" which is some code we assume has multiple predecessors, may or may not come from a clearly-separated copy of the pre-optimized program.

@o-, your discussion of the scoping of invalidate is somewhat overlapping with the question of how to articulate (1) and (2) together (the purpose of the landing pad being to translate from the pre-invalidation environment in (1) to the post-invalidation environment of (3), it should logically be in charge of transitioning between distinct scopes, but our IR so far has no clear representation of "scope transition".

You propose to have more logic in the invalidate instruction proper (do renaming at this level, and possibly re-definition of folded constants), as a consequence that makes the landing pad a bit less useful. Initially when we discussed having more logic in invalidation, there was some pushback (from Jan I think, hoping for a simple instruction), and we decided to use landing pads instead as a temporary measure. With your experience implementing optimizations, we are seeing more drawbacks of the current approach. I think we could ask the following questions:
A. what is precisely the problem that requires a redesign of the scope?
B. is the proposed solution something that can be expressed as normal instructions and scope annotations (in which case maybe it could remain in the landing pad?) or requires logic not currently expressible in the language (in which case we could extend the "invalidate" instruction to support it, or the language itself)?

(I won't say anything more specific right now because I need to look more at your useful examples of scoping issues.)

The `end_opt` instruction is, on the other hand, related to the boundary between (2) and (3), I think. I think it is an interesting idea, but on the other hand I find it a bit strange that we need to say this explicitly: in my mind, (3) naturally starts when we reach code that has multiple predecessors, and in which it is thus, in the general case, difficult to assume anything -- I would have assumed that our optimizations would naturally stop at this step. Maybe we would need to be able to tell our optimizers about "entry points", that is functions that may be called from outside the module and for which optimizations that rely on knowing all the predecessors are thus impossible. This is a sort of dual to your proposal ("exit points", in a sense), but maybe it is a bit more natural and useful in the general case? Of course, your proposal, while being a bit more ad-hoc, has the advantage of obviously doing the right thing for invalidation-specific optimizations, while "the optimizer will stop here because one of the predecessors itself flows from an entry point" is an emergent property that we would need to check for each optimization (but is this not part of correctness checking?).

o-:
@gasche it's very subtle. If you think about scope as a dataflow property, then:

In your first listing on line 1 d is not in scope, since the incoming edges are the program entry point with scope {} and the goto with scope { d } (the goto does not have the empty scope outgoing!)

In your last listing on line 8 d is in scope since it has only one incoming edge where d is in scope.

If I reorder your last listing the trace for the deopt case looks like this:

      const d = 1
      goto deopt_entry_loop2
    deopt_entry_loop2:
      goto deopt_cont_loop2
    deopt_cont_loop2:
      print d
      goto loop.0
    loop.0:
      const d = 1
      branch (d==1) cont.0 loop2.0
    loop2.0:
      print d
      goto loop.0
    cont.0:
      stop

now it's obvious that there is in fact shadowing.

o-:
If you want the outgoing scope of the goto on line 6 to be empty, then we need a new definition for scopes! E.g. we need to say that jumps eliminate dead variables. Or what kind of definition do you have in mind?

But I am not sure this solves all the problems.

Also it means that the out-scope of branch depends on the branch taken. That seems a bit unpleasant.

gasche:
The way I think of scopes, there are no "outgoing scopes" -- indeed, if we had ongoing scopes, there would be one for each successor of the instruction, instead of just one for the instruction; just like, in Hoare Logic, an expression with multiple return points will have one postcondition for each return point. There are ingoing scopes, and the "outgoing scopes" of an instruction are the ingoing scopes of its successors.

Then the question of whether the `loop.0` label has an empty ingoing scope (as I chose in my rendering of the program) or `{d}` in scope (as you point out the dataflow analysis could determine) is possibly a bit subjective: it depends on whether we consider that scopes are an integral part of the program text, only it can be omitted by the user and then inferred by the compiler, or whether they are an analysis property that the user does not control. I rather have the first interpretation in mind (and this is one reason why I pushed to let users write explicit scope annotations). With this interpretation, it seems natural (to me) that `loop.0`, being obtained by doing a fresh copy of `loop`, would have the exact same entering scope -- if you had asked me to guess naively what the branch pruning optimization does, I would have assumed that it preserves the scopes as it copies (but in fact it only copies the instructions, the scopes are not available at this moment in the implementation).

Note that whether `d` is inferred in scope at this point by the scope analysis depends on how we implement it. If we have a forward analysis, as implemented today, then yes. But we could have a bacwkard analysis that would add variables in scope as they are used, and remove them as they are defined, and then it would not be inferred in `loop.0`'s scope. Intuitively one could expect the inferred scope to be as tight as possible, for example to be the intersection of the forward and backward inferred scopes, but I am not sure whether just taking the intersection is actually correct. (Clearly there should be an error if the forward defined-variables analysis does not include a variable mentioned in the backward require-variable analysis; we already do an equivalent check in the `finish` part of `Scope.infer`.).

o-:
Ok, thanks for explaining. I think I understand your point better now.

You are right in the sense that I see the scope as purely inferred property and the annotations as just confirmation to the user that the scopes are what he thinks they are.

The reason I took on this pov is that I find it very hard to reason about them otherwise. A very simple example where I already got stuck:

    1     const a = 1
    2     print a
    3  {} const b = exp

What happens if I move line 3 above 2? Do I take the attached scope with me (then I break the scope of print a) or do I leave the scope there (then I potentially turn an illegal program into a legal one if exp consumes some variables which become available).

It seems that scopes are more "global properties" and it seems strange to associate them with instructions. If anything I could see a variant where the basic block entries are annotated. Or how do you see this?

So my pov was that the scope is basically a model for the stack frame of the function. This corresponds with our invalidate being described as a scope transformation (because OSR transforms one stack frame into another one).

I like your idea of tightening the scope! Doing the inference backwards should fix the conflict I found. Also I believe the inferred scopes would grow/shrink like slots of a stack machine in that case.

gasche:
Moving instructions with their scope together may be easier in the "small-step" style, where you update your scope for each instruction you pass. If we see how to do it in small-step, we could also find out how to do it in big-step style: if you move inside the def-use chain of a variable, then you should add it to your scope?

We should consider having an explicit `kill a, b, c` instruction that would remove variables from scope. Then it is very clear what the liveness intervals are. I mentioned this a while ago, but initially I thought it would be a bit heavy (that inferred scopes would be lighter for the same information). But those would force us to be more careful when we move instructions around, and also help us do the right thing. (I agree with your mental model of the lexical environment as a stack shape and the relation between stack replacement and substitutions.)

(The `kill` instruction may also help clarifying what is happening in a landing pad / invalidate. A substitution `v2 := v1` could be expressed as `{const,mut} v2 = v1; kill v1` in the landing pad.)

o-:
> Moving instructions with their scope together may be easier in the "small-step" style

But my example is small step style.... I want to move instruction 3 up one step. Still I don't see how to accomplish it. You would say add a to the scope? That's only valid if a is not in exp. And in general I think there is not one answer. its basically a design choice. Whether we say to the user the scope is anchored at this point in the control flow or at this instruction. I still find both confusing :)

> We should consider having an explicit kill

what happens if my scope includes a killed an a non-killed version?

      const d
      branch exp a b
    a:
      kill d
    b:
      print d

in general, having un-balanced branches is tricky! (think of the stack machine which has to manage a consistent stack height)

gasche:
> You would say add a to the scope? That's only valid if a is not in exp

Yes, I would add `a` to the scope. I would also assume that the input of the optimization pass is scope-correct: we are only operating on scope-correct programs and preserving this property. I'm not worried about "invalid programs becoming valid".

I think that if we want to make our variable-manipulation problems go away, and having explicit `kill` instructions is part of this (I have written `kill` enough times by now that I would support rather using a more pacific keyword for this, for example `undecl` or `forget` or `drop`; I like `drop`), I think that we should be very explicit and strict in the scoping rules, by removing the ability to implicitly drop variables on a transition (the "less than equal to" we wrote on paper when we first worked out the scoping analysis), and instead require that the ingoing scope of each instruction be *exactly equal* to the outgoing scope of all its successors. With that point of view in mind, your example would be statically rejected, as the predecessors of `b`, namely `branch` and `drop`, have different scopes.

That can have an important impact on the bytecode representation of existing programs that were more flexible about scope. I resisted suggesting this previously, because the current, more implicit system is more convenient for the human reader or writer of code, but I now wonder if we haven't been optimizing the wrong thing: we should try to maximize the convenience of code transformations, if that is what we want to study.

I think that the typical painful case in this more strict approach is the jump in a big scope into a label into a smaller scope:

    {b,c} branch e l1 {b}l2

This needs to be rewritten

    {b,c} branch e l1 {b,c}l2'
    l2':
      kill b
      {b} goto {b}l2

But my intuition would be that we will *not* have to implement this rewriting as a program transformation: assuming that we only accept input programs that respect this more stringent scoping requirements, we should be able to conveniently transform them while preserving this requirement. So once the transition is made (if we decide to make it) this becomes a non-problem of the past.

o-:
Can we maybe take a step back? We are not converging. I would propose to start from the requirements and how our discussions stem from them.

## Variables

We have two types of variables in sourir. const and mut.

1. The idea with const was to capture runtime constants. Therefore you can only assign once to it.
2. Mut are the normal assignable variables.

Regarding 1) How exactly are we planning to use the const variables? Should they have an ssa like property such that we can always point to their unique definition or do we accept conflicting definitions, ie do we accept:

    if exp then const a = 2 else const a = 3; print a

Is this relaxed version strong enough? My intuition is that it's probably ok but it does not buy us anything. Because you can rewrite to

    mut a'; if exp then a' <- 2 else a' <- 3; const a <- a'; print a

Therefore I think we should disallow it.

Regarding 2) For mutable variables its always safe to maximize the scope (ie. start the program and start the landing pads with all the required declarations). In this case we do not have to worry about any scopes when doing transformations.

## Scopes

We see scopes as our model of the frame of the function.

1. In the optimized output the scopes have to be as tight as possible (ie. correspond with the liveness intervals) because otherwise its not a reasonable model of a frame
2. All incoming control flow must agree on a scope because otherwise you cannot map it to a static frame.
3. In the invalidate case there must be a well defined mapping between optimized scope and unoptimized scope. Whether that mapping is just a filter operation or potentially something more powerful is an open question

o-:
Now regarding our discussion so far I observe:
* If we accept that a const declaration should always strictly dominate its uses then imho no trickery with scopes will solve the initial problem. Line 13 in your optimized program is not strictly dominated by line 9! If on the other hand we accept that even a const might have two conflicting definitions then I don't see a really good reason to keep it in the language and not just use mut everywhere.
* Scope inference should really proceed backwards! we do not care if both outgoing branches look the same. We care about the merging points in control flow.
* If we want scopes to converge with liveness we need the drop instruction. Alternatively we say the function frame is the live variables in scope. But then I'd say scopes where way to prominent in the design.
* I still fail to see the requirement/use case for making the scopes part of the language

o-:
Btw. there is some work on implementing OSR for the llvm IR.
(http://www.dis.uniroma1.it/~delia/papers/cgo2016-osrkit.pdf, http://www.sable.mcgill.ca/publications/papers/2013-1/vee13-lameed.pdf)

There you can see how in the context of an ssa ir the osr-entry point is a phi-node for all live variables.

o-:
Actually I just realized that if we say that declarations must not just dominate the uses but be in a dominating **block** our problems vanish! It won't matter how we define when a scope ends. The scope will always start early enough to dominate the whole landing pad.
update: but then const is less expressive

gasche:
What invalidation needs to be able to do, to be recognized as "serious invalidation", is to jump to a point in the un-optimized program that is *in the middle* of things. I think this means that "a variable declaration must dominate its uses" is too strong a requirement; to jump "in the middle", in particular I want to jump between a declaration and its use, which means the invalidation instruction (or its landing pad) must be able to also declare the variable, and you have two declarations flowing to the same site. This is precisely what is happening in the program we are discussing.

If we were using a SSA form, there would be a phi-node at `loop2.0`, merging the declarations of `d` flowing from 7 and 10. Would you feel more comfortable with such a design?

o-:
> What invalidation needs to be able to do, to be recognized as "serious invalidation", is to jump to a point in the un-optimized program that is in the middle of things.

I fully agree. More and more I think we should keep the old program as is and figure out how to jump back to it.

> If we were using a SSA form, there would be a phi-node at loop2.0, merging the declarations of d flowing from 7 and 10. Would you feel more comfortable with such a design?

This really depends on what we want to be able to do with a const variable! I am not sure myself...

gasche:
In my mind, const variables are meant to refer to previous computation results (in the style of variables of A-normal form), be local, and be very fast to create and discard. They correspond to stack slots in a stack-based VM, registers in a register-based VM, and SSA variables in a SSA IR. On the other hand, mut variables correspond to a heavier language-level concept that is less efficient to create and discard, and introduces mutability.

In a SSA IR, a join point can merge two independent declaration of a variable with a phi-node. In a stack IR, I think it makes sense to enforce local join points to have all predecessors with the same stack height. This corresponds to allow distinct declarations to be joined in the resulting scope, but to require the same variables to be bound, or at least the scope of each successors to be a superset of the joined scope.

I am convinced that we need to allow this (separate definitions to flow to the same point) for both categories of variables to properly model invalidation.

Re. having `drop` vs. the more implicit scopes, I personally have no strong preference, but the fact that scoping proves a delicate part of the design makes me want to go for the most explicit solution, that is explicit drop and very strict scope requirements. I'll try to implement `drop` in the following days, so that we can look at a concrete proposal.

o-:
> I am convinced that we need to allow this (separate definitions to flow to the same point) for both categories of variables to properly model invalidation.

Yes I convinced myself of the same.

I think we'll manage without explicit ssa

> Re. having drop vs. the more implicit scopes

I tried to formalize scope checking (without annotations) and I think we do not need it. See: #41 
If we say the scoping is valid if **all required** variables are in scope we avoid having to talk about what happens about unused variables.


On a related note I played around with this changes to the semantics, which imho makes most problems we discussed in this thread go away: #41

In particular every program in the compilation unit will be scoped independently. With this new scheme the branch pruning will be:

```
# Version 1
 1 { } loop:
 2 { }   const d = 1
 3 {d}   branch (d==1) cont loop2

 4 {d}  loop2:
 5 {d}   print d
 6 {d}   goto loop

 7 { } cont:
```

to

```
# Version 2
 1 { } loop:
 2 { }   const d = 1
 3 {d}   osr (d!=1) "Version 1" loop2 [d = d]

# Version 1
 1 { } loop:
 2 { }   const d = 1
 3 {d}   branch (d==1) cont loop2

 4 {d}  loop2:
 5 {d}   print d
 6 {d}   goto loop

 7 { } cont:
```

The condition for osr to be valid in respect to scoping would be: everything which is required after the osr-entry must be set-up by the osr instruction. See #41

o-:
PS: I am fine with dumping this PR as it is

o-:
to be rewritten after the new semantics from #41 are implemented

PR #41, opened 2017-02-01
Semantics changes

o-:
this new version brings:
* multiple versions of the same program
* osr instruction which jumps between versions
* a formal definition of scopes

### DIFF
--- a/doc/ir-semantics.text
+++ b/doc/ir-semantics.text
@@ -117,7 +117,8 @@ reach_decl P l x =
 
 valid_at_l P l =
   ∀ x ∈ requires P(l) :
-    reach_decl P l x
+    ∀ l' ∈ pred P l :
+      reach_decl P l' x
 
 valid_at P =
   ∀ l ∈ P : valid_at_l P l
--- a/doc/ir-semantics.text
+++ b/doc/ir-semantics.text
@@ -3,21 +3,21 @@
 x        variables
 
 l        labels
+v        optimization version
 
 a        addresses
 
+C ::= (v ↦ P)*          compilation unit
 P ::= (l ↦ i)*          program
 
 i ::=    instructions
-| const x = e           constant variable
-| mut x = e             mutable variable
-| x <- e                assignment
-| branch x l₁ l₂        conditional
-| goto l                goto
-| print x               print
-| invalidate(e, l, x̄)   invalidate
-
-x̄        list of variables
+| const x = e               constant variable
+| mut x = e                 mutable variable
+| x <- e                    assignment
+| branch x l₁ l₂            conditional
+| goto l                    goto
+| print x                   print
+| osr(e, v, l, (x = exp)*)  osr : (can be used for osr-in and osr-out)
 
 e ::=    (simple) expressions
 | lit                   literals
@@ -40,7 +40,7 @@ E ::= (x ↦const v | x ↦var a)*  lexical environment
 H ::= (l ↦ v)                   mutable heap
 
 T ::= (lit)*                    output trace
-Configuration: (T, H, E, P, l)
+Configuration: (T, H, E, C, P, l)
 
 Lookup (partial) function, returns a v:
   (H,E)[x] :=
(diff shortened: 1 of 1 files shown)


doc/ir-semantics.text
```diff
@@ -58,30 +58,74 @@ Evaluation of simple expressions:
...
+     l'
+     when P = {..., l -> _, l' -> _, ...}
+
+pred P l =
+  { l' ∈ P : succ P l' = l }
+
+reach_decl P l x =
+  (x ∈ declares P(l)) ∨
+  ∀ l' ∈ pred P l :
+    reach_decl P l' x
```
o-: ugh that is undefined for invalid scopes. Not sure yet how to fix

gasche:
Regarding the multi-versions: it's an interesting idea, but I would like to make sure that it is necessary before going on with it. The reason is that I believe that some of the problems we had may occur for some other transformations than branch pruning, and it may not be the case then that "but we are creating a separate/different version" will be an answer that makes sense then. So I would like to understand whether this is an ad-hoc solution, a general one.

In terms of scopes, if I understand correctly, the main effect of the version is to introduce a new starting point for each version that is "grounded" to be at an empty scope. But another option we would have is to have a `start` instruction, the dual of `stop`, that would represent an entry point for the module, and be grounded to have the empty scope. Would that also solve our problem? Is it better in some way? (It seems a bit more flexible to me, but I'm not yet fully informed.)

In term of name conflicts, do we allow two versions to use the same variable for different purposes without conflict?

o-:
> it's an interesting idea, but I would like to make sure that it is necessary before going on with it.

It's probably not necessary as in its the only way of doing it. But it is very close to the actual thing we are trying to model + solves the problems.

> some of the problems we had may occur for some other transformations than branch pruning, and it may not be the case then that "but we are creating a separate/different version" will be an answer that makes sense then.

Lets categorize our optimizations into speculative and non-speculative. For speculative optimizations we use osr (invalidate) instructions. Roughly speaking we want to remove instructions from the program which might still be used. It's possible since they are still in the old version. For non-speculative optimizations they are basically traditional always-correct transformations. They should keep the program correct without additional tricks. That should be off-the shelf compiler stuff. If we encounter problems there we need to fix the semantics of the existing instructions.

> In terms of scopes, if I understand correctly, the main effect of the version is to introduce a new starting point for each version

plus it naturally limits the scope of transformations to one version

> But another option we would have is to have a start instruction

that sounds more ad-hoc to me. the start instruction would only be used for scope inference, right? because there might be more than one of them.

> In term of name conflicts, do we allow two versions to use the same variable for different purposes without conflict?

yes they have nothing in common. The osr instruction is the only way of getting from one version to the other.

gasche:
After giving it more thought, I like the way that the new "version" structure correlates more closely to the way we think about OSR. You actually call them "modules" in the patch, which also makes sense. In fact, we are giving more structures to the IR (the very first version had functions, this is somewhat like that), in order to better map to existing systems or transformation narratives. (The fact that we don't have to worry about naming conflicts across modules is also nice.)

The one thing I don't really like in the patch are the scoping rules; they are too "big-step"! You don't really define "scoping", you define a first-order proposition that describes "what it means to be well-scoped". I would rather have a scoping judgment that annotates each instruction with a scope, and check that it is consistent with successors or predecessors -- closer to what we have now in the implementation, and as you sketch at the end of your patch. For example, we could define scoping as a mapping from each instruction to a set of variable (the ingoing scope) such that:
- if the instruction has no predecessor (if we have `start` instructions, if it is `start`) the scope is empty
- all the variables used (free) in the instruction are included in the scope
- the scope of all the instruction's successors is smaller than the scope plus the variables defined (bound) in the instruction (if we have `drop` instructions, we can ask for equality instead of superset; and then we have to add "minus the variables dropped" here).

I would propose the following:
1. we merge the current PR and implement modules (a program is a `(module * instr array) list`?)
2. we decide whether to introduce explicit `drop` and `start` instructions (I still think that both are nice, `drop` as it simplifies scoping and clarifies the runtime meaning of scopes, and `start` as the dual of `stop`; I would propose to implement a check that all modules start with a `start`, that they don't have free codes not reachable from a `start` position, and that all code paths end on an `osr` or `stop`)
3. depending on (2)'s decision, we update the description of scoping

o-:
> if we have start instructions, if it is start

I think we want a start label! (see the super nice updated scoping rules)

gonna respond to the rest later, first coffee

doc/ir-semantics.text
```diff
@@ -58,30 +73,71 @@ Evaluation of simple expressions:
...
-    when P(l) = invalidate(e, l', x̄)
+  (T, H, E, C, P, l) -> (T, H, E, C, P, succ l)
+    when P(l) = osr(e, _, _)
+     and (eval H E e) is not (true)
+
+  (T, H, E, C, P, l) -> (T, H, E', C, P', l')
+    when P(l) = osr(e, v, l', (x = ee)*)
      and (eval H E e) is (true)
+     and P' := C(v)
+     and E' := (x ↦ ⊥ | (eval H E ee))*
```
gasche: What is it that you mean here by `(x ↦ ⊥ | (eval H E ee))*`? The question is the what the domain of those `x` is: if it is those that appear in `(x = ee)*`, then the `⊥` case does not really make sense as they are all defined. I suppose you have the expected scope of `l'` in `C(v)` in mind, but we should specify that more precisely.

Another choice would be to take a list of the form `(mut x (= ee)? | const x = ee)` in argument, that would really create the whole lexical environment (all bindings would have to be listed, even undefined one), and clearly distinguish `mut` and `const` as this level. I think that would be nicer, what do you think?

o-: since eval is not defined for ⊥ I wanted to say that e.g. (x = ⊥, y = 1+2) maps to the environment {x -> ⊥, y -> eval 1+2}

but yeah the notation could be improved.

adding the mut/const modifiers sounds like a good idea, though note that the target scope should fully define whether they are const or mut

gasche:
I think that `start` should be an instruction rather than a label, to be symmetric with `stop` (which has to be an instruction, otherwise the successors of stop labels would be defined as special cases, which is ugly). A start label is the just the label of a `start` instruction.

Your new scoping judgment is nicer, indeed. One thing that I find still a bit unsatisfying is that it is an algorithm rather than a specification: it enforces the "minimal scopes" property (compared to the previous definition that would not specify that, but only require that required variable be available and allow to drop variable on forward transitions), and conversely some properties that hold of this judgment are not obvious to check, they are require a proof on the implementation (for example, the property that for each instruction, all its required variables will be in scope). Maybe we could formulate a scoping specification, as a judgment, and a scoping inference algorithm, also as a judgment -- it is useful to have both.

Finally, I think the names and variables `C` and `P` are backwards. A Program should be a *set* of Compilation Units, and not the other way around.

o-:
> I think that start should be an instruction rather than a label

the advantage of the label is, that the start label is unique and at the beginning by construction (see line 13). In the case of a start instruction we would need to enforce well-placedness and uniqueness externally. But since I do not have a strong opinion on this you can decide.

> Your new scoping judgment is nicer, indeed.

One thing I am not sure: the derivations have circular branches. Think of the program:

```
1: x = 1
2:
3: branch x 2 4
4:
```

The derivation for 2 is:

```
----------
∅ ⊨ 1             x ⊨ 3       <---\
-----------------------             |
x ⊨ 2                               |
----------                          |
x ⊨ 3           -------------------/
```

do you find this a problem?

>  it enforces the "minimal scopes" property

That's true. but on the other hand they are well defined. In particular this allows us to very precisely specify what it means for the osr-instruction to set up a correct scope.

> it is useful to have both

I agree. How about we say (1) is the property of a program to be well-scoped and (2) the other is the derivation of the Frame.

> Finally, I think the names and variables ... are backwards

agree

Let me quickly address those things.

Are you interested in implementing it?

gasche:
I was thinking that we could have several start instructions in the program (the idea came before you proposed to have separation units, but in fact it still makes sense within a compilation unit, although it is not necessary anymore). Currently we have one start and several stops, but there is no reason for the asymmetry. If we enforce that the only instruction without successors are `stop`, and the only instructions without predecessors are `start`, we have a nicely symmetric system.

> One thing I am not sure: the derivations have circular branches. [..] Do you find this a problem?

Yes and no. I think we could avoid this by changing the judgment to take the scope decisions (mapping from (version, pc) to a scope) as an input, then on labels we only need to check that the label's scope is correct, without needing to recursively re-check scoping from the label's instruction. I am planning to formulate the "specification" judgment in that style after you make the minor P/C swap.

> Are you interested in implementing it?

Sure, why not.

Where are we on the question of having a instruction flow graph (a graph where nodes are instructions instead of blocks)? I am not completely sure that we need to materialize this as an object separate from the source code, or whether we could just keep using the predecessors/successors to implement traversals from the source code.

o-:
> Currently we have one start and several stops, but there is no reason for the asymmetry.

ok, but there is also no reason for multiple starts, is there? How about removing stop instead? We can say the module is: M ::= s -> i, (l -> i)*, e -> ⊥  and stop will become "goto e"

> I am planning to formulate the "specification" judgment in that style after you make the minor P/C swap

Ok, you are good to go :)
I already added the scop vs. frame versions, but if you can make it non-recursive that would be great!

>  I am not completely sure that we need to materialize this as an object separate from the source code, or whether we could just keep using the predecessors/successors to implement traversals from the source code.

Yeah that might indeed be possible

o-:
> I already added the scop vs. frame versions, but if you can make it non-recursive that would be great!

actually I just broke it, because the new  ⊨(scp) thingy happily allows shadowing :(
In the ⊨(fra) case it is prevented because you have to remove what you declare when going up, but in the (scp) version you can leave it in the set and then remove it on the second declaration.

But I'll leave it to you for now...

gasche:
Thanks. I'll be on a train for a good part of this afternoon, so I may push back late or tomorrow morning.

o-:
awesome, have a good ride

PR #42, opened 2017-02-05
acyclic scoping judgments, with comments

gasche:
This is a follow-up on #41. @o-, what do you think?

(I don't have super-strong opinions on the specific syntax of the judgments, in particular where parameters are placed. I just tried something a bit different because it seemed interesting. Feel free to suggest to change back on any on which you have a well-defined preference.)

### DIFF
--- a/doc/ir-semantics.text
+++ b/doc/ir-semantics.text
@@ -121,49 +121,83 @@ declares i =
 requires i =
   | mut x = e
   | const x = e
-  | osr e, _        -> vars(e)
+  | osr (e, v, l, (x = ee)*)
+                    -> vars(e) ∪ vars(ee)*
   | x <- e          -> x :: vars(e)
   | branch x _
   | print x         -> [x]
   | _               -> []
 
-update S i = S \ declares i ∪ requires i
+scope S ::= x*
+scope assignment A ::= ((v, l) → S)*
 
-## Well-scoped module
+## Declarative scoping judgment
 
-M, S ⊨(scp) l
--------------   ∀ l ∈ M : ∃ S
-     ⊨(scp) P
+This judgment just classifies "possible" scope assignments A for
+a program P. It does not characterize the tightest possible
+assignment¹, as it allows to implicitly drop variables from scope on
+each transition (the ⊇ relations in the (v, l → i) rule). In
+particular, it cannot forbid shadowing, as it is always possible to
+drop all same-named bindings before a declaration.
 
-M, S_l* ⊨(scp) l*
------------------   ∀ l* ∈ pred (M, l) : ∃ S_l* ⊂ update S
-   M, S ⊨(scp) l
+¹: the tightest possible assignment can be defined impredicatively as
+the pointwise intersection of all valid judgments, which is
+well-defined given that valid scopes are bounded by the set of
+variables defined in the program.
+
+  ∀(v → M) ∈ P,   A, P ⊨ v → M
+  ----------------------------
+            A, P ⊨
+
+  ∀(l → i) ∈ M,   A, P ⊨ v, l → i
+  -------------------------------
+         A, P ⊨ v → M
+
+  S := A(v,l)
+  requires i ⊆ S
+  ∀l* ∈ pred(P(v),l), A(v,l*) ⊇ S
+  ∀l' ∈ succ(P(v),l), S ∪ declares i ⊇ A(v,l')
+  A, P ⊨ i : S
+  ----------------------------------------
+  A, P ⊨ v, l → i
+
+The judgment (A, P ⊢ i : S) below is the instruction-specific judgment
+that checks the scope restrictions that are specific to each
+instruction, rather than the general control-flow constraints and the
+generic requires/declares handling.
+
+  ----------------
+  A, P ⊨ start : ∅
+
+              A(v, l) = x*
+  ----------------------------------
+  A, P ⊨ osr(e, v, l, (x = ee*)) : S
+
+  ---------------
+  A, P ⊨ stop : ∅
 
 
--------------
-M, ∅ ⊨(scp) s
+## Algorithmic scoping judgment
 
+(A, P ⊨), (A, P ⊨ v → M), (A, P ⊢ i : S): as in the declarative
+scoping judgment. Only (A, P ⊨ v, l → i) differs.
 
-## Frame construction
+  S := A(v,l) \ declares i
+  ∀l* ∈ pred(P(v),l), A(v,l*) ⊇ S ∪ requires i
+  A, P ⊨ i : S
+  --------------------------------------------
+  A, P ⊨ v, l → i
 
-⊨(fra) M    P ⊨(osr) M
-----------------------   ∀ M ∈ P
-       ⊨(fra) P
+There is a fundamental difference in the role of the scope assignment
+A between the declarative and algorithmic judgments.
 
+In the declarative judgment, A(v,l) points to the *ingoing* scope
+of the instruction at location (v,l): the scope that is available when
+arriving to the instruction from a predecessor. The scope available to
+the successors is computed from this value as (A(v,l) ∪ declares i).
 
-M, S ⊨(fra) l
--------------   ∀ l ∈ M : ∃ S
-     ⊨(fra) P
-
-M, update S ⊨(fra) l*
----------------------   ∀ l* ∈ pred (M, l)
-       M, S ⊨(fra) l
-
-
--------------
-M, ∅ ⊨(fra) s
-
-
-P(v), x̄ ⊨(fra) l
-----------------   ∀ osr(e, v, l, (x = ee)*) ∈ M
-      P ⊨(osr) M
+In the algorithmic judgment, S := A(v,l) points to the *outgoing*
+scope of the instruction at location (v,l): the scope that is avaiable
+when leaving the instruction to a successor. The scope visible on
+entering the instruction is (A(v,l) \ declares i) *if* shadowing is
+forbidden (otherwise it may contain any subset of the (declares i)).
--- a/doc/ir-semantics.text
+++ b/doc/ir-semantics.text
@@ -183,24 +183,41 @@ generic requires/declares handling.
 scoping judgment. Only (A, P ⊨a v, l → i) differs.
 
   S := A(v,l) \ declares i
-  ∀l* ∈ pred(P(v),l), A(v,l*) ⊇ S ∪ requires i
+  (⋂_{l* ∈ pred(P(v),l)} A(v,l*)) = S ∪ requires i
   A, P ⊨a i : S
-  ----------------------------------------
+  ------------------------------------------------
   A, P ⊨a v, l → i
 
 There is a fundamental difference in the role of the scope assignment
 A between the declarative and algorithmic judgments.
 
-In the declarative judgment, A(v,l) points to the *ingoing* scope
-of the instruction at location (v,l): the scope that is available when
-arriving to the instruction from a predecessor. The scope available to
-the successors is computed from this value as (A(v,l) ∪ declares i).
+In the declarative judgment, A(v,l) points to the variables
+*available* in the *ingoing* scope of the instruction at location
+(v,l): the scope that is available when arriving to the instruction
+from a predecessor. The variables that are available in the outgoing
+scope can be computed from this value as (A(v,l) ∪ declares i).
 
-In the algorithmic judgment, S := A(v,l) points to the *outgoing*
-scope of the instruction at location (v,l): the scope that is avaiable
-when leaving the instruction to a successor. By asking that the scope
-visible on entering the instruction be (A(v,l) \ declares i), we force
-variables to be dropped before they are re-declared: shadowing is
-technically possible, but inserting `drop` instructions (on each (⊃)
-link where scopes are not (=)) would clarify that no shadowing
-actually happens.
+In the algorithmic judgment, A(v,l) points to the variables that are
+*required* in the *outgoing* scope of the instruction at location
+(v,l): the scope that is required when leaving the instruction to
+a successor. The variables required in the ingoing scope can be
+computed from this value as (A(v,l) \ declares i).
+
+In the declarative judgment, the relation between an instruction scope
+and its neighbors is under-specified: a predecessor's outgoing scope
+may have more variables than our ingoing scope, and any successor's
+ingoing scope may have less variable than our outgoing scope.
+
+In the algorithmic judgment, the relation between the scopes of an
+instruction and its neighbors is deterministic: an instruction's
+ingoing scope is exactly the variables that are common among all its
+predecessors. Implicitly dropping a variable that will never be used
+again is possible: this is done by choosing a A(v,l) that is smaller
+than the instruction's ingoing set plus (declares i).
+
+Problem: I think there is a problem with the notion of outgoing scope
+as it is currently defined for multi-successor instructions, as it
+forces all successors to see the same set of required
+variables. I think that introducing `drop` instructions would be the
+cleanest way to work around these issues – by removing any
+underspecification from the declarative specification.
--- a/doc/ir-semantics.text
+++ b/doc/ir-semantics.text
@@ -154,9 +154,9 @@ variables defined in the program.
          A, P ⊨d v → M
 
   requires i ⊆ A(v,l)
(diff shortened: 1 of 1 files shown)


doc/ir-semantics.text
```diff
@@ -121,49 +121,103 @@ declares i =
...
+ingoing scope may have less variable than our outgoing scope.
+
+In the algorithmic judgment, the relation between the scopes of an
+instruction and its neighbors is deterministic: an instruction's
+ingoing scope is exactly the variables that are common among all its
+predecessors. Implicitly dropping a variable that will never be used
+again is possible: this is done by choosing a A(v,l) that is smaller
+than the instruction's ingoing set plus (declares i).
+
+Problem: I think there is a problem with the notion of outgoing scope
```
o-, line 220: I don't understand this comment. I don't see any outgoing scopes defined. S is the incomming scope, not?

e.g. you can have:

      { } mut a = 3
      {a} branch a x y
    x:
      {a} print a
    y:
      { } stop

ie. not both outgoing scopes of the branch are the same, one is {} the other is {a}.

Imho this should be compatible with both of your judgments

doc/ir-semantics.text
```diff
@@ -121,49 +121,103 @@ declares i =
...
+
+  S := A(v,l)
+  requires i ⊆ S
+  ∀l* ∈ pred(P(v),l), A(v,l*) ⊇ S
+  ∀l' ∈ succ(P(v),l), S ∪ declares i ⊇ A(v,l')
+  A, P ⊨d i : S
+  ----------------------------------------
+  A, P ⊨d v, l → i
+
+The judgment (A, P ⊢ i : S) below is the instruction-specific judgment
```
o-: you call it  ⊨d  elsewhere

doc/ir-semantics.text
```diff
@@ -121,49 +121,103 @@ declares i =
...
+
+  ∀(l → i) ∈ M,   A, P ⊨d v, l → i
+  -------------------------------
+         A, P ⊨d v → M
+
+  S := A(v,l)
+  requires i ⊆ S
+  ∀l* ∈ pred(P(v),l), A(v,l*) ⊇ S
+  ∀l' ∈ succ(P(v),l), S ∪ declares i ⊇ A(v,l')
+  A, P ⊨d i : S
```
o-: its a bit hard to see how S is bound from both sides. I think part of the problem is hiding the l parameter S consumes. how about we define

    S(l) ::= A(v,l)

and write:

    <forall> l' <elemnt> pred :  S(l*) ⊇ S(l)

doc/ir-semantics.text
```diff
@@ -121,49 +121,103 @@ declares i =
...
+## Declarative scoping judgment
+
+This judgment just classifies "possible" scope assignments A for
+a program P. It does not characterize the tightest possible
+assignment¹, as it allows to implicitly drop variables from scope on
+each transition (the ⊇ relations in the (v, l → i) rule). In
+particular, it cannot forbid shadowing, as it is always possible to
+drop all same-named bindings before a declaration.
+
+¹: the tightest possible assignment can be defined impredicatively as
```
o-, line 143: I very much like this definition

doc/ir-semantics.text
```diff
@@ -121,49 +121,103 @@ declares i =
...
+scope S ::= x*
+scope assignment A ::= ((v, l) → S)*
+
+## Declarative scoping judgment
+
+This judgment just classifies "possible" scope assignments A for
+a program P. It does not characterize the tightest possible
+assignment¹, as it allows to implicitly drop variables from scope on
+each transition (the ⊇ relations in the (v, l → i) rule). In
+particular, it cannot forbid shadowing, as it is always possible to
```
o-: why can't it? let's just require:

    ∀l* ∈ pred(P(v),l), A(v,l*)  <intersection> (declares i) is empty

doc/ir-semantics.text
```diff
@@ -121,49 +121,103 @@ declares i =
...
+drop all same-named bindings before a declaration.
+
+¹: the tightest possible assignment can be defined impredicatively as
+the pointwise intersection of all valid judgments, which is
+well-defined given that valid scopes are bounded by the set of
+variables defined in the program.
+
+  ∀(v → M) ∈ P,   A, P ⊨d v → M
```
o-: syntax is all fine, but please ":" and not "," :)

doc/ir-semantics.text
```diff
@@ -121,49 +121,103 @@ declares i =
...
+
+  ----------------
+  A, P ⊨d start : ∅
+
+              A(v, l) = x*
+  ----------------------------------
+  A, P ⊨d osr(e, v, l, (x = ee*)) : S
+
+  ---------------
+  A, P ⊨d stop : ∅
```
o-: is this one needed?

doc/ir-semantics.text
```diff
@@ -121,49 +121,103 @@ declares i =
...
+that checks the scope restrictions that are specific to each
+instruction, rather than the general control-flow constraints and the
+generic requires/declares handling.
+
+  ----------------
+  A, P ⊨d start : ∅
+
+              A(v, l) = x*
+  ----------------------------------
+  A, P ⊨d osr(e, v, l, (x = ee*)) : S
```
o-: wait, this one is wrong. S refers to the wrong l,v. it should be:

             A(v', l') = x*
    ----------------------------------
    A, P ⊨d v, l -> osr(e, v', l', (x = ee)*) : A(v,l)

o-: also the (*) has to come outside of the (x = ee)

o-: Note: this places some non-local relations on the scopes from different modules. That's why I fear it will be hard to speak about correctness if the scopes are under-defined

o- (changes requested):

o-:
In general it looks good. I have two highlevel things:

1. I would like to see the rules for declarative/algorithmic to be more similar (e.g. only substituting a relation). This would make it more clear how they relate.
2. I think we could specify the conditions on A in a more declarative style. Let me elaborate in a separate comment:

o-:
```
↥l  ::=  declares(P(v, l))                   (declared vars)
↧l  ::=  requires(P(v, l))                   (minimally required vars)
↘l  ::=  A(v, l)                             (scope)
↗l  ::=  A(v, l) ∪ ↥l                        (outgoing scope)

l ⥴ l'  ⇔  l' ∈ succ P(v, l)                (a preorder)


A  with  l* ⥴ l  ⇒  ↗l* ⊇ ↘l ⊇ ↧l           (scope of l: in out scope of l* and big enough)
---------------------------------
A, P ⊨d v, l → i
```

Update:

ok I see how the precise version does not work when the var is used only on one branch, in which case we indeed need to say something like: its used on at least one successor (which you accomplish with the intersection) :(

Update2:

well, you had the good definition. I think we need to look at all predecessors. I can't simplify it :) in my notation I come up with:

```
↘l  =  ⋂ {↘l' : l' ⥴ l} \ ↥l
```

doc/ir-semantics.text
o-: Update: Are you referring to this sentence:

> In the algorithmic judgment, A(v,l) points to the variables that are *required* in the *outgoing* scope of the instruction at location

In that case I think the fix is to say "variables [...] are *required* in " **at least one of** " the *outgoing* scope **s**

o-:
After our discussion on the phone I would really like to simplify the problem, because I think it distracts us. My new proposal is:

* we have one scope judgment
* it is declarative
* scope is strictly growing, you cannot drop anything
* it prevents shadowing

the rule is:

```
A  with  l* ⥴ l  ⇒  ↗l*  =  ↘l  ⊇  ↧l
---------------------------------
A, P ⊨d v, l → i
```

or:

```
A  with  l* ⥴ l  ⇒  scope(l*) ∪ declared(l*) =  scope(l) ⊇  used(l)
------------------------------------------
A, P ⊨d v, l → i
```

Note: this disallows jumping above a declaration => less trouble for us...

For any well-scoped program we can infer the livenes of variables. This is only interesting for the implementation as it allows it to set them to ⊥ in the osr case (thus not inflate their liveness). But this is the concern of the optimistic optimization and not of the scope design.

also, I am not really pushing for this silly notation. It was just fun to try.

gasche:
If we cannot have a variable declaration in a loop with a backedge, does your last proposal not correspond essentially to having all variables global?

doc/ir-semantics.text
gasche: It is not strictly needed, but (1) it gives the nice property that algorithmic and declarative share all rules but one (it is more important in the algorithmic case), and (2) it does not reduce expressivity in any way, as we can always pick `A(v,l) = ∅` whenever `v, l -> stop` -- that is, implicitly drop all variables just before going to `stop`, which is sensible anyway.

doc/ir-semantics.text
gasche: I don't think that the rule as written is wrong, as the judgment is of the form `A, P ⊨d i : S` and not `A, P ⊨d v, l -> i : S`. But indeed my first version used `v', l'` as variable names, and maybe that was less confusing. I will restore that.

I think it makes sense to have the osr checking depend on the scopes of the other modules, but indeed it means that it is more natural to define a dynamic semantics for a program `P` *paired* with its scope assignment `A`. The explicit `drop` that I am now proposing are designed to weaved this information directly in the code.

doc/ir-semantics.text
gasche: Good catch, thanks.

doc/ir-semantics.text
gasche: The problem is with a small variable of the example of code you gave:

```
  { } mut a = 3
  {a} branch a x y
x:
  {a} print a
  stop
y:
  { } stop
```

it is compatible with the declarative judgment, but not with the algorithmic judgment: the scope of `y: { } stop` is wrong according to the rule that the scope here should be *exactly* the intersection of the scopes of all predecessors: the only predecessor here is `{a} branch`, so the scope must be `{a}` -- which means this judgment is inadequate to capture minimal scoping.

I think indeed that to fix this I would need to have, for each instruction, one scope per successor (so a set of annotations of the form `A(v,l,l')` in the algorithmic case). I think this starts to get very heavy and rather painful to reason about, which is what decided me to go the explicit `drop` route, where we would write instead:
```
  { } mut a = 3
  {a} branch a x y
x:
  {a} print a
  drop a
  stop
y:
  {a} drop a
  stop
```

doc/ir-semantics.text
o-: > as the judgment is of the form A, P ⊨d i : S and not A, P ⊨d v, l -> i : S

Ok wait I am really confused now!

Just to make sure if I understand. Your Idea is that in the following program:

```
1: start
2: mut a = 1
3: print a
```

line 3 would be proven by:

```
A ::= {1 -> [], 2 -> [], 3 -> [a]}


    ----------------
     A, P ⊨d start : ∅
(1) -----------------
     A, P ⊨d mut a = 1 : ∅
(2) -----------------
     A, P ⊨d print a : [a]
(3) -----------------------------------------------------------
     A, P ⊨d v, 3 → print a
```

There is currently no rule for step (1) and (2). Also it would not work to have (1) and (2) not carry up the labels, because otherwise in the program:
```
1: start
2: print a
3: mut a = 1
4: print a
```
I could prove line 4 by using the A of line 2!

doc/ir-semantics.text
o-: Right, both our algorithmic judgment do less than what our implementation did.

Actually it makes sense, because the only way it could be minimal is by requiring both start and end to have an empty scope (and not like now: either of them).

Should we remove the algorithmic version (both) for now? They seem broken.

gasche: > Should we remove the algorithmic version (both) for now? They seem broken.

Yes, we can do that. (I think it was very helpful to have written them down as support for the discussion.)

doc/ir-semantics.text
gasche: The judgment is not a tree in this way. The rule

```
  ∀(l → i) ∈ M:   A, P ⊨d v, l → i
  -------------------------------
         A, P ⊨d v → M
```

does a check on each instruction (and each check looks at all predecessors and successors, but not recursively), and then each of those checks also does a `A, P ⊨ i : A(v,l)` instruction-specific check. So the proof tree for that program would be as follows:

```
P :=
  1: start
  2: mut a = 1
  3: print a

A := 1 -> {}, 2 -> {}, 3 -> {a}

∅ ⊆ A(1)              ∅ ⊆ A(2)                  {a} ⊆ {a}
(pred(1) is ∅)        A(1) ⊇ A(2)               A(2) ⊇ A(3) (wrong!)
A(1) ∪ ∅ ⊇ A(2)       A(2) ∪ {a} ⊇ A(3)         (succ(3) is ∅)
A, P ⊨ start : ∅      A, P ⊨ mut a : ∅          A, P ⊨ print a : {a}
----------------      --------------------      -------------------
A, P ⊨ 1 → start      A, P ⊨ 2 → mut a = 1      A, P ⊨ 3 -> print a
-------------------------------------------------------------------
A, P ⊨
```

There is an issue indeed with the "wrong!" part, I need to add a `declares i` somewhere, will look after lunch.

o-: Aha, but then you need to specify that

A, P ⊨d i : A(v,l)

is a root in your proof. So far it looks like only "start" "stop" and "osr" are. Also You should probably require that i is not "start" or "stop" because otherwise the following A is correct too:
```
A := 1 -> {x}, 2 -> {x}, 3 -> {x,a}
```

Also in the osr case you have to require the special osr rule to be true as well as the normal scope to exist!

o-: as far as the (wrong!) goes I thing the correct equation would be:
```
A(v,l*) ⊇ A(v,l) ∪ declares i ⊇ A(v,l')
```

Update:
Actually:
```
A(v,l*) ∪ declares P(v, l*) ⊇ A(v,l)     and
A(v,l) ∪ declares P(v, l) ⊇ A(v,l')
```
which is probably over-specified!

o-: What's kind of confusing to me is the following: Your judgment form should actually be

```
  --------------------            (<- note this additional bar)
  [...]
  A, P ⊨d i : A(v,l)
  --------------------
  A, P ⊨d v, l → i
```

And now if you try to write that down in the standard way of only requiring one step at a time you can't. Because in

```
----------------
A, P ⊨d i : S
```

You lost the label l on the right hand side and can (ab)use the scope from another identical instruction

gasche: > You lost the label l on the right hand side and can (ab)use the scope from another identical instruction

I don't think there actually is any issue with this, in the sense that the `A, P |= i : S` judgment does not actually depend on the label of the `i` instruction (in the `osr` case it is only used to check that the scope of the osred-to destination is identical to `S`). But I see that this is troubling you, so I would be perfectly happy to write that judgment `A, P |= v, l -> i : S` instead. I started this way, then removed the location as I realized it was not necessary.

gasche: > as far as the (wrong!) goes I thing the correct equation would be: [..]

Yeah, I had the same thought path as you just did, so in the new version (I just pushed a next commit, d063ede55d94f8d277fddcbdfd5056531a5b0df8) I removed the `pred` requirement. I think actually just check the successors is enough: the predecessors are checked by their own predecessors.

On the other hand, I realized that I had not imposed any starting condition on judgment without predecessors that are not `start`. (In my mind there is a separate check that ensures that all instructions without predecessors are `start`, and all instructions without successors are `stop`, but we have not agreed on that and it is not particularly important so for now I'm lax about it, maybe we'll change the formalization back to having designated start *labels*). So I added this as an explicit condition to this rule : `pred(P(v), l) = \emptyset \implies i = start`.

o-: > But I see that this is troubling you

It's ok, I get it now. It was really the missing rule that threw me off completely. Because naturally in my head all your proof trees had to have the only ground rule "start" at the top :)

o-: > On the other hand, I realized that I had not imposed any starting condition on judgment without predecessors that are not start

Now with ```i ∉ {start, osr, stop}``` its indeed only a problem for starts, which are no start.

o- (approved):

gasche:
Thanks for the thorough review! I merged.

PR #43, opened 2017-02-06
Explicit drop

gasche:
@o- This is the PR that I referred to during our video call today -- but I had forgotten to submit it! It's only a first step.


o- (approved):
lgtm

PR #44, opened 2017-02-07
get rid of the assembler

o-:
It's getting annoying to maintain assembler and parser in parallel


gasche:
But we had so much fun with the subtyping! lgtm.

o-:
its so sad ;)

PR #45, opened 2017-02-07
keep track of mut/const in scope inference

o-:


gasche:
Are the `_untyped` variants a transitional thing? I would support getting rid of "untyped" variables all over the place, and writing `const x`, or `mut x` instead of just `x`. Did you plan to do this later, or is the current compromise to stay?

o-:
I think we still need the untyped variants, don't we?

e.g. when I ask what are the used variables at "print x" I don't know if its mut or const

the untyped variants are used in those cases. e.g. when you need to check that the used variables of "print x" are contained in the current scope.

gasche:
That sounds fishy to me. I would rather have `required : env -> instr -> TypedVarSet.t`, with an exception on undeclared variable happening in this function. (Sure, that does make a purely backward analysis of scopes difficult, but I think we should rather do a two-pass forward then backward to compute the minimal scopes anyway.) But with that understanding of the pain point, I would be fine with merging now and evolving things later -- especially if you build off this on the next PR.

o-:
I agree in spirit. But as an example if you look at

https://github.com/reactorlabs/sourir/pull/45/files#diff-4c856ffb7eaeaaa40a4d7716044e4849R51

when we intersect with the annotations this would be hard to do without the untyped version.

o-:
maybe we should really start thinking about having a more structured representation of the program internally. Eg. some typed ssa for the variables...

PR #46, opened 2017-02-07
implement segments and new osr

o-:
implement new osr according to the formal semantics

needs #45


o-:
@gasche this PR is not yet finished. I need some help with the parser.messages. Can you tell me what I should do?

eval.ml
```diff
@@ -199,34 +200,46 @@ let reduce conf =
...
+           (Env.add x (Const (eval conf e)) env,
+            heap)
+         | OsrMut (x, OsrExp e) ->
+           let a = Address.fresh () in
+           (Env.add x (Mut a) env,
+            Heap.add a (Value (eval conf e)) heap)
+         | OsrMut (x, OsrUndef) ->
+           let a = Address.fresh () in
+           (Env.add x (Mut a) env,
+            Heap.add a Undefined conf.heap)
```
o-: what we didn't realize in the formal version is that creating new bindings extends the heap... (even if eval is pure)

disasm.ml
```diff
@@ -33,16 +33,36 @@ let disassemble_annotated (prog : annotated_program) =
...
+      dump_expr exp;
+      pr buf " %s %s [" v l;
+      let last = (List.length vars) - 1 in
+      List.iteri (fun i x ->
+          begin match x with
+            | OsrConst (x, e) -> pr buf "const %s = " x; dump_expr e;
+            | OsrMut (x, OsrExp e) -> pr buf "mut %s = " x; dump_expr e;
+            | OsrMut (x, OsrUndef) -> pr buf "mut %s" x
+          end;
+          if (i < last) then pr buf ", "
```
o-: is there a nice idiom for this?

transform.ml
```diff
@@ -1,157 +1,78 @@
...
+          let osr = List.map (function
+              | Mut_var x -> OsrMut (x, OsrExp (Simple (Var x)))
+              | Const_var x -> OsrConst (x, (Simple (Var x))))
+              (TypedVarSet.elements scope) in
+          branch_prune pc'
+            (Goto l2 :: Osr (exp, deopt_label, l1, osr) :: acc_i)
+            (None :: a :: acc_a)
+        | i, a ->
+          branch_prune pc' (i::acc_i) (a::acc_a)
+        end
```
o-, line 81: check out how simple branch prune is now :)

eval.ml
gasche: Having a heap that is never garbage-collected is ok-ish. (In particular there is never any garbage collection at this level.) Note that `drop`-ing a mutable variable properly shrinks the heap, so this problem will go away with strict dropping rules.

disasm.ml
gasche: Not that I know of. `String.concat : string -> string list -> string list` takes a delimiter, but this is in a different style. One way to avoid calling `List.length` is to define the inner `match` as its own function (dump_osr_decl)`, and then do

```ocaml
begin match vars with
| [] -> ()
| hd::tl -> dump_osr_decl hd; List.iter (fun x -> pr buf ", "; dump_osr_decl x) tl
end
```

transform.ml
```diff
@@ -1,157 +1,78 @@
...
-            begin match dead_code.(pc') with
-            | None -> remove_dead_code pc'
-            | Some _ -> Comment c :: remove_dead_code pc'
-            end
-      | None, _ -> remove_dead_code pc'
-      | Some _, _ -> prog.(pc) :: remove_dead_code pc'
+    if pc = Array.length instrs then
+      (Array.of_list (List.rev acc_i), Array.of_list (List.rev acc_a))
+    else
+      match[@warning "-4"] unreachable_code.(pc), instrs.(pc), annots.(pc) with
```
gasche: You can probably drop the `[@warning "-4"]` here, as we are not matching on any head constructor anymore (it used to be `Commet`). (If it starts warning, then my guess was wrong and it needs to remain.)

gasche:
> Can you tell me what I should do?

`make update-parser-messages` should update the `parser.message` file in place (ugh!) such that `git diff` (or `meld parser.messages` show you the impact that the changes to the grammar had on the error state. You may have new error paths to existing error states with a manually-written message (in which case the message may need to be updated), and you will have new error states that do not have an error message yet.

In this specific case, I just tried from your branch, there are two subtleties:
1. It does not work very well because in the present commit you just removed `parser.messages`; you should un-remove this file before doing this, because menhir will recognize the old `parser.messages` file and try to preserve its information (instead of asking you to redefine all error messages anew).
2. If you have the old `parser.messages` around, it will fail because this file refers to the token name `INVALIDATE`, which does not exist anymore. So before running `update-parser-messages` you should search and replace all occurrences of `INVALIDATE` by `OSR` in `parser.messages` (also replacing the lowercase occurrences in the error messages would make sense).

To someome that is not super-familiar with parser generators, what happens next is not obvious. I can offer to do the work of updating the file (it should not be very hard for me), but if you would like to learn, we could also use github for this: if you managed to do the two things above, you could create a commit with the resulting state of `parser.messages`, submit a PR about it, and I could comment on it to give you hints to iteratively rebase the PR.

eval.ml
o-: Yes, what I meant was that before the osr did not touch the heap but just the env. Now it does create new bindings.

o-:
thanks for the help. I think I found out how to fix the messages

o-:
btw. regarding the disassembly output:

the best solution imho would be to allow [mut x,] (ie. trailing comma).

but I did not manage do get an unambiguous grammar which supports this...

eval.ml
gasche: Oh. Would it be possible, for the variables that you repurpose instead of defining afresh, to reuse the previous heap address? I think of allocation as a more expansive operation than, say, `const` declaration, so reuse would make sense.

o-: I am not raising this for performance reasons.

Manual object pooling like you suggest might of course be possible in some cases but I think letting the hypothetical GC deal with it is just fine. Note: its not always possible. You can create a new value e.g. "[mut a = 3+2]".

I was more concerned for the proof. It's a globally visible effect that our old invalidate did not have.

o-:
@gasche this should be ready now

eval.ml
gasche: This is not an optimization, I believe it is important semantically -- it might be relevant for the proof as you also suggest. If two versions have *identical* instruction streams, I want to be able to insert an `osr` at some place that jumps to the same place in the other version, and this should be a no-op (besides the version change). This only works if you are able to say, like we merge two different constant declarations (morally, we have a phi-node), that we merge two different notions of mutable variable into one (and in particular, the heap cells are shared as well). Allocation on the osr also makes perfect sense semantically, but only in the case where the corresponding allocation was avoided in the incoming version in the first place, for example by sinking a mutable variable into a constant variable.

o-: ok, that's what I suspected.

Should I restrict the osr again?

I could for example support:

```
const x = expression | mut x = y | mut x
```

without allocation

gasche: We could do a clear distinction between renamings and new declarations (in terms of the existing scope). We could for example have `x <- y`, which requires that both variables be of the same mutability class (what you call "type"), and then the general declaration forms `const x = e`, `mut x = e` and `mut y`, that are understood to declare new variables.

o-: yeah, this might be interesting in the future. But for all we need osr now it's very easy to get back to the non-allocating version by not having expressions for mutable variables.

let's do the simple thing first

PR #52, opened 2017-02-10
add travis integration

o-:
travis ocaml infrastructure due to mirage/mirage

### DIFF
--- a/tests.ml
+++ b/tests.ml
@@ -567,6 +567,10 @@ let suite =
    ]
 ;;
 
-let _ =
-  run_test_tt_main suite;
-;;
+let () =
+  let res = run_test_tt_main suite in
+  if (not (List.for_all (fun r ->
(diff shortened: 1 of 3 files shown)


tests.ml
```diff
@@ -567,6 +567,10 @@ let suite =
...
 
-let _ =
-  run_test_tt_main suite;
-;;
+let () =
+  let res = run_test_tt_main suite in
+  if (not (List.for_all (fun r ->
+      match r with
+      | RSuccess _ -> true
+      | _ -> false) res)) then
```
gasche: Not very readable.

```
let is_success = function
  | RSuccess _ -> true
  | _ -> false in
if not (List.for_all is_success test_result) then exit 1;;

gasche:
Very nice! I was too lazy to do this for too long. Merge whenever.

PR #53, opened 2017-02-10
Moded variables

gasche:
These two commits do not change the implementation behavior, they are purely internal changes. I initially intended to get rid of `*_untyped` as a further commit in the branch, but I am not sure anymore that this is the best move, so I just kept these.

### DIFF
--- a/instr.ml
+++ b/instr.ml
@@ -105,12 +105,12 @@ module VarSet = Set.Make(Variable)
 
 type variable_type = Mut_var | Const_var
 
-type typed_var = variable_type * variable
+type moded_var = variable_type * variable
 
 exception Incomparable
 
-module TypedVar = struct
-  type t = typed_var
+module ModedVar = struct
+  type t = moded_var
   let compare (ma, a) (mb, b) =
     match String.compare a b with
     | 0 ->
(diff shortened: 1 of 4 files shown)


instr.ml
```diff
@@ -103,59 +103,36 @@ let resolve (code : instruction_stream) (label : string) =
...
-    | Mut_var   a, Const_var b
-    | Const_var a, Mut_var   b ->
-      if a = b then raise Incomparable else String.compare a b
+module ModedVar = struct
+  type t = moded_var
+  let compare (ma, a) (mb, b) =
+    match String.compare a b with
+    | 0 ->
+      if ma = mb then 0 else raise Incomparable
+    | c -> c
```
o-, line 118: I would argue that comparing (Mut_var a) with (Const_var a) should also raise an error since there should not be two equally named variables with different mode in scope

instr.ml
```diff
@@ -103,59 +103,36 @@ let resolve (code : instruction_stream) (label : string) =
...
+  let vars set = List.map snd (elements set)
   let untyped set = VarSet.of_list (vars set)
 
   let muts set =
-    let muts = filter (fun v ->
-        match v with
-        | Mut_var x -> true
-        | Const_var x -> false)
-      set in
+    let muts = filter (fun (m,v) -> m = Mut_var) set in
```
o-, line 128: oh my. If I only knew ocaml :) thanks

o- (approved):
Except for the small nitpick it lgtm

instr.ml
gasche: I think this is what the code does: `compare` returns `0` when the two inputs are the same, and then we check that their mode are the same otherwise we fail. I tried to follow your (reasonable) behavior on this.

o-: my bad, of course

PR #54, opened 2017-02-10
Strict scope

gasche:


gasche:
Most of the simplifying action, which is a first start on the simplicity benefits I expect with this approach, is in the [transform.ml changes](https://github.com/reactorlabs/sourir/pull/54/commits/7e4db6489587df9117a2edb6f42e66bff8c7a2a7#diff-e87230a5806e9927be124ca6f5a2893c) in 7e4db6489587df9117a2edb6f42e66bff8c7a2a7.

sourir.ml
```diff
@@ -16,33 +16,60 @@ let () =
...
         exit 1
-      | exception Scope.DuplicateVariable (xs, pc) ->
+      | Scope.ExtraneousVariable (xs, pc) ->
         let l = pc+1 in
         begin match Instr.VarSet.elements xs with
-          | [x] -> Printf.eprintf "%d : Error: Variable %s is declared more than once.\n%!" l x
-          | xs -> Printf.eprintf "%d : Error: Variables {%s} are declared more than once.\n%!"
-                    l (String.concat ", " xs)
+          | [x] -> Printf.eprintf
+                     "%s%d : Error: Variable %s is unexpected in scope.\n%!"
```
o-: I think this error message is a bit confusing. The most user visible appearance of scopes are the scope annotations. Therefore its more like the variable is unexpectedly **not** in scope (-annotations). I would say something like "Variable %s not mentioned in scope {%s}"

o-:
Random observation: We cannot assign a scope to
```
loop:
  const a = 1
  branch a==1 loop cont
cont:
```
Because there is no space for a "drop a" after the branch. But on can write:
```
loop:
  const a = 1
  branch a==1 loop' cont
loop':
  drop a
  goto loop
cont:
```
But it's cumbersome

o- (approved):
This lgtm.

Wheren't you arguing for making the annotations more part of the program. That's why I recently changed transform.ml to preserve the annotation when copying code. I more like the idea of annotations being a verification that the scopes are what the user thinks they are. Therefore naturally I support this PR. I assume your change of mind comes from the fact that the fine grained control of scope is now possible by drop. Therefore the feature of "shrinking scope by smaller annotation" is not needed anymore. Is that a correct characterization?

gasche:
I agree with your random observation, and in fact it is in the commit message of 324a00f3b33935e355fe8eaede12a91bd4bd2635.

Having annotations part of transform.ml was also a way to fix some of the examples that we discussed previously, such as the one where the copied program would start at a different scope than the initial program (this one is fixed with versions). By forcing annotations to be perfectly inferrable from the code, we remove that difficulty. But yes, it's kind of a heavyweight approach.

sourir.ml
gasche: Fixed in b8adddee6d76b4465672dc7c74059f02786d8fc6, thanks.

PR #57, opened 2017-02-14
major cleanup of analysis code

o-:
just cosmetics. making analysis api more uniform.

### DIFF
--- a/analysis.ml
+++ b/analysis.ml
@@ -1,34 +1,44 @@
 open Instr
 
-let successors (instrs : instruction_stream) pc =
+let pcs (instrs : instruction_stream) : pc array =
+  Array.mapi (fun pc _ -> pc) instrs
+
+let successors_at (instrs : instruction_stream) pc : pc list =
   let pc' = pc + 1 in
-  let next = if pc' = Array.length instrs then [] else [pc'] in
+  let instr = instrs.(pc) in
   let resolve = Instr.resolve instrs in
-  match instrs.(pc) with
-  | Decl_const _
-  | Decl_mut _
-  | Assign _
-  | Drop _
-  | Clear _
-  | Read _
-  | Label _
-  | Comment _
-  | Osr _
-  | Print _ -> next
-  | Goto l -> [resolve l]
-  | Branch (_e, l1, l2) -> [resolve l1; resolve l2]
+  match instr with
+  | Decl_const _ | Decl_mut _ | Assign _ | Drop _ | Clear _ | Read _
+  | Label _ | Comment _ | Osr _ | Print _ ->
+    let is_last = pc' = Array.length instrs in
+    if is_last then [] else [pc']
   | Stop -> []
+  | Goto l -> [resolve l]
+  | Branch (_e, l1, l2) when l1 = l2 -> [resolve l1]
+  | Branch (_e, l1, l2) -> [resolve l1; resolve l2]
 
-let predecessors (instrs : instruction_stream) =
-  let preds = Array.map (fun _ -> []) instrs in
-  let mark_successor pc pc' =
-    preds.(pc') <- pc :: preds.(pc') in
-  for pc = 0 to Array.length instrs - 1 do
-    List.iter (mark_successor pc) (successors instrs pc)
-  done;
-  preds
+let successors (instrs : instruction_stream) : pc list array =
+  let succs_at pc = successors_at instrs pc in
+  Array.map succs_at (pcs instrs)
 
-let dataflow_analysis (next : pc -> pc list)
+let predecessors (instrs : instruction_stream) : pc list array =
+  let succs = successors instrs in
+  let pcs = pcs instrs in
+  let preds_at pc =
+    let is_pred_at pc_2 pc_1 = List.exists ((=) pc_2) succs.(pc_1) in
+    List.filter (is_pred_at pc) (Array.to_list pcs)
+  in
+  Array.map preds_at pcs
+
+let starts (instrs : instruction_stream) = [0]
+
+let stops (instrs : instruction_stream) =
+  let succs = successors instrs in
+  let is_exit pc = succs.(pc) = [] in
+  let pcs = Array.to_list (pcs instrs) in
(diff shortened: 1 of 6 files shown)


analysis.ml
```diff
@@ -1,34 +1,44 @@
 open Instr
 
-let successors (instrs : instruction_stream) pc =
+let pcs (instrs : instruction_stream) : pc array =
+  Array.mapi (fun pc _ -> pc) instrs
```
gasche, line 4: This reminds me that in some place you increment pc by 1 before reporting them to the user, to be in the range 1..N instead of 0..(N-1), and I think maybe I forgot to do this in some other. Maybe we could just keep the 0..(N-1) range for simplicity, given that we already gave up on a perfect matching between pc and line number?

analysis.ml
```diff
@@ -1,34 +1,44 @@
...
-  | Osr _
-  | Print _ -> next
+  match instr with
+  | Decl_const _ | Decl_mut _ | Assign _ | Drop _ | Clear _ | Read _
+  | Label _ | Comment _ | Osr _ | Print _ ->
+    let is_last = pc' = Array.length instrs in
+    if is_last then [] else [pc']
+  | Stop -> []
   | Goto l -> [resolve l]
+  | Branch (_e, l1, l2) when l1 = l2 -> [resolve l1]
```
gasche: I think this is a bit too clever -- I don't want to have to remember to do this whenever we add new instructions. Could we have a `uniq` function somewhere that we would call?

Locally `uniq li` can be defined as `li |> PcSet.of_list |> PcSet.elements` -- but we could also have a generic definition on lists, in a `lib.ml` module for general libraries, or depend on a library that already defines this (eg. [containers](https://github.com/c-cube/ocaml-containers), [batteries](https://github.com/ocaml-batteries-team/batteries-included) or JaneStreet's [base](https://github.com/janestreet/base)).

analysis.ml
```diff
@@ -1,34 +1,44 @@
...
-    preds.(pc') <- pc :: preds.(pc') in
-  for pc = 0 to Array.length instrs - 1 do
-    List.iter (mark_successor pc) (successors instrs pc)
-  done;
-  preds
+let successors (instrs : instruction_stream) : pc list array =
+  let succs_at pc = successors_at instrs pc in
+  Array.map succs_at (pcs instrs)
+
+let predecessors (instrs : instruction_stream) : pc list array =
```
gasche, line 29: For most programming language features and style there are several stages to learning, one of them being "finding out how/when to use it" and a later one "finding out how/when not to use it". You seem to be somewhere in the middle :-) Your code for this function is exactly as long as the previous one, but it is harder to understand (more higher-order functions) and quadratic instead of linear.

analysis.ml
```diff
@@ -48,42 +58,40 @@ let dataflow_analysis (next : pc -> pc list)
...
-  let init_state = List.map (fun pc -> (init_state, pc)) exits in
-  let preds = predecessors instrs in
-  let predecessors pc = preds.(pc) in
-  dataflow_analysis predecessors init_state instrs
+let forward_analysis init_state seg merge update =
+  let successors = successors seg in
+  let starts = starts seg in
+  assert (starts != []);
+  let init = List.map (fun pos -> (init_state, pos)) starts in
+  make_total (dataflow_analysis successors init seg merge update)
```
gasche, line 87: One thing I considered while reading this code is to actually make `starts` return the set of instructions without predecessors, dually to `stops`. This would give the guarantee that the analysis is well-defined on all pcs.Then, we could have a `make_total` function that would fail with an exception (or even an assertion failure) at analysis time, rather than at analysis-query time:

```
let make_total result =
    let of_option = function
    | None -> assert false
    | Some res -> res
    in
    let result = Array.map of_option result in
    fun pc -> result.(pc)
```

and further simplify the caller's code.

I am not completely sure this whole suggestion is a good idea. (But decoupling analyses in general from dead-code elimination could be a reasonable choice.)

analysis.ml
o-: the pc still is 0...(N-1). This helper just creates an increasing integer array the same size as the instruction stream. It's there since I did not find a convenient ocaml one-liner for creating a sequence. Do you know one?

gasche: I realize that here it's still 0..(N-1), but I think that you use `+1` before displaying pcs in error messages -- this is what I was referring to.

Your one-liner is fine, I think it is actually the best you can do. Another one is `Array.init (Array.length instrs) (fun i -> i)`. Some extended standard libraries (I mention a couple ones below) have other functions, but I think `mapi` is appropriate here.

analysis.ml
o-: That's what I had first. I can switch back.

gasche: I would recommend it: less clever code is good.

analysis.ml
o-: oh, I misunderstood the old version. I mainly wanted to make it dependent on successor (which it already was :) ).

analysis.ml
o-: > One thing I considered while reading this code is to actually make starts return the set of instructions without predecessors, dually to stops.

That won't fly as it is now. It is entirely possible to have a program start with a label thus the entry does have a predecessor. Also you can have dead code without a predecessor.

> Then, we could have a make_total function that would fail with an exception

I think we should not require all code to be reachable for the analysis to be correct. That's very annoying if you need to figure something out in the middle of rewriting code. Therefore I think it makes sense to only error if you query the status of unreachable code without anticipating for unreachable code.

analysis.ml
o-: Btw. you probably mean in the pretty printer / scope inference. There you talk about line numbers but they are 0-based. In some cases you pretty-print the code with the pc, so its not that bad. But in other cases (eg. when the annotation is too small) you just say: "n : Error: Variable x is not declared." and n is a pc, but not a line number.

analysis.ml
gasche: Ok, we can keep it in mind but any change here would be for later.

PR #58, opened 2017-02-15
Codemotion reloaded

o-:
Simple code moving transformation:

Moving assignments of the form ```x <- exp``` as far up the callgraph as possible.

Since we are not in SSA moving assignments is only possible (without further analysis) if the assignments dominates all its uses. Otherwise we might accidentally capture uses on some traces.

The condition for a move to be valid is that the move target dominates the move origin (ie. we are moving upwards) and is dominated by all reaching defs (ie. we are not moving above our dependencies).

We only look at our own use-def chain. Thus the transformation renames the variable to avoid overriding unrelated uses of the same name. For example:

```
mut x = 1
print x
x <- 2
print x
```

Becomes:
```
mut x0 = 2
mut x = 1
print x
print x0
```


gasche:
@ulysses4ever we could pair-review this, ping me when you have time.

tests.ml
```diff
@@ -502,13 +457,61 @@ jmp:
...
+      end:
+       stop
+      bla:
+       const z = 1
+       mut x = 1
+       mut y = z
+       goto loop
+  " in
+  let res = Transform.hoist_assignment t in
+  assert_equal (main res).(8) (Decl_mut ("y0", Some (Simple (Var "z"))));
```
ulysses4ever: Is it possible to see the whole opt'ed program?

rename.ml
```diff
@@ -0,0 +1,74 @@
+open Instr
+
+let fresh_var instrs var =
+  let cand i = var ^ (string_of_int i) in
```
ulysses4ever, line 4: [gasche] I would propose that `cand "x" 0` be `"x"` instead of `"x0"` to have the nice sequence x,x1,x2...

tests.ml
```diff
@@ -502,13 +457,61 @@ jmp:
...
+       mut x = 1
+       mut y = 2
+      loop:
+       y <- x
+       x <- (x + y)
+       branch (x==10) end loop
+      end:
+  " in
+  let res = Transform.hoist_assignment t in
+  assert_equal (main res).(3) (main t).(3);
```
mhyee: It would also be nice to make it more explicit that in this case (and the next one), code motion can't be applied. Maybe in a comment that also explains why?

transform.ml
```diff
@@ -78,3 +78,53 @@ let branch_prune (prog : program) : program =
...
+  let rec move pc =
+    if pc != to_pc then begin
+      instrs.(pc) <- instrs.(pc+dir);
+      move (pc+dir)
+    end
+  in
+  move from_pc;
+  instrs.(to_pc) <- from
+
+let hoist_assignment prog =
```
mhyee: I would also like a comment that explains code motion and its conditions. Something like the text of the PR is good enough.

mhyee:
Looks good. @gasche, @ulysses4ever, and I had some small comments, but once they're resolved, I think this can be merged in.

gasche:
Thinking about this again, I'm not completely sure that I understand why a separate mutable variable (with a fresh name) is needed. Would reusing the same variable be wrong? Would using a constant variable be wrong?

o-:
> Would reusing the same variable be wrong?

let me give you an example

```
1   mut a
2   a <- 1
3   print a
4   a <- 2
5   print a
```

Line 4 dominates all uses therefore we can move it up. We want to move it as far up as possible. Therefore we place it below its latest reaching definition. Since it uses nothing we can move it arbitrarily up. If we rename then that's all there is to it.

Now, if we do not rename there are more things to consider:
1. We cannot move above the declaration (ie. we move below line 1)  -> simple enough
2. It's not enough to look at reaching definitions. But we also need some notion of "interfering statements". For example we would additionally need to state that we do not move above an instruction which defines or uses the same variable.

Or to give you a more sensible example, we could not lift this assignment:

```
   mut x
   read x
   mut a
loop:
    a <- x
    print a
    a <- a + 1
    print a
    goto loop
```

> Would using a constant variable be wrong?

I think it should work since we effectively take a slice where the variable is assigned only once.

gasche:
Thanks for the extra examples, this is very useful. I'll open a ticket for making the new declaration a constant (it makes sense in terms of the imaginary performance model), it's a perfect "junior job" for people getting to get involved in the implementation.

PR #59, opened 2017-02-15
Scope uninitialized

o-:
Initially I just wanted to fix a warning (see 336a468).

But then I realized that scope inference and checking for uninitialized variables really should be split into two independent steps

### DIFF
--- a/scope.ml
+++ b/scope.ml
@@ -51,7 +51,7 @@ exception IncompatibleScope of inference_state * inference_state * pc
 let infer instructions : inferred_scope array =
   let open Analysis in
   let merge pc cur incom =
-    if not (ScopeInfo.equal cur.info incom.info)
+    if not (ModedVarSet.equal cur.info.declared incom.info.declared)
     then raise (IncompatibleScope (cur, incom, pc))
     else if PcSet.equal cur.sources incom.sources then None
     else Some { info = cur.info; sources = PcSet.union cur.sources incom.sources }
@@ -160,7 +160,6 @@ let explain_incompatible_scope outchan s1 s2 pc =
           "  - the %s %s %a and the %s does not\n"
           name1 verb print_vars diff name2 in
     print_diff "declares" diff.declared;
-    print_diff "defines"  (ModedVarSet.diff diff.defined diff.declared);
   in
   Printf.bprintf buf
     "At instruction %d,\n\
--- a/scope.ml
+++ b/scope.ml
@@ -32,7 +32,6 @@ let drop_annots : annotated_program -> program =
   List.map (fun (name, (instrs, annot)) -> (name, instrs))
 
 exception IncompatibleScope of inference_state * inference_state * pc
-exception SometimesUninitialized of inference_state * inference_state * pc
 
 let infer instructions : inferred_scope array =
   let open Analysis in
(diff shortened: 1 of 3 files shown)


o-:
Update: @gasche did you intend to not allow a merge between a undefined/defined variable?

Consider e.g.

```
  mut x = 1
  branch (x==1) clearit skip
clearit:
  clear x
skip:
  x <- 1
  print x
```

This was illegal at pc 4 since it merges x in defined and undefined state. Imho this is too conservative (also our scoping judments never forbid this) and the changes in 0d790c6 make it legal again

gasche:
This was indeed my intention: I thought that it would be consistent with the scoping rule, given that I also added an explicit `clear x` instruction to make mutable variable undefined again.

If you think that relaxing this constraint is better, we can go for it. It was my intuition, however, that the stricter constraint would also be useful for optimization purposes: knowing for sure that every parent defined the value seems like a good invariant to have. (It also has the convenient consequence that we completely rule out "undefined value" errors at runtime, instead of having just a heuristic warning.)

o-:
> This was indeed my intention

I see. I don't have a strong opinion to be honest. I just wanted the warning not to talk about scopes when the problem was that a variable is maybe undefined. Since it confused even me.

> It also has the convenient consequence that we completely rule out "undefined value" errors at runtime, instead of having just a heuristic warning.

That is not true. The proposed chage is less conservative. But it's still conservative. It won't allow use of an undefined variable.

o-:
I think I will revert to your intended behavior but keep the two analysis separate and improve the wording of the error message. How about that?

gasche:
> The proposed chage is less conservative. But it's still conservative. It won't allow use of an undefined variable.

Oh, I assumed you merged (Uninit + Init -> Init), of course if you merge (-> Uninit) then it's fine, sorry.

I am fine with a different error message. What are the benefits of splitting the two analyses, though? It creates some amount of code duplication. Could have the better message with a single-pass computation still? I'm just asking of curiosity, I have no strong opinion; the factoring of the two computations (and the Scope_info) type came from you anyway, so you are welcome to take it back if you prefer ;-)

o-:
> What are the benefits of splitting the two analyses, though?

It seems to create accidental complexity because the two things do not really belong together. There is not much code duplication since the common stuff is in forward_analysis. If I remove the scope_info code gets shorter (even with the duplicated lines). But I guess it's a matter of taste and I am not even sure myself :)

gasche:
I propose then to honor your work by merging the PR.

At some point we might want to have some form of integration tests that also check the error messages. (I like the fact that having this makes it easy to review error-message changes in PRs as they are present in full in reference files for tests.)

o-:
I am totally unsure whether the transformations in #58 respect those additional scoping rules! One argument against them would be that moving code around does have to respect two different scoping rules (actual scopes and uninitialized-scopes).

Ie. without this restrictions you can:

1. maximize scope
2. move code
3. shrink scope

this is what I more or less do in #58 (at least 1&2) by converting assignment into declaration.

With the restriction (2) has to respect the uninitialized-scopes. I have to think more about what this entails.

o-:
Lets assume the following example:

```
  mut a
  branch exp b1 b2
b1:
  a <- 1
  print a
  goto cont
b2:
  mut b = 1
  a <- b
  print a
cont:
```

Now if we perform the assignment hoisting from #58 the following will happen:

```
  mut a0 <- 1
  mut a
  branch exp b1 b2
b1:
  print a0
  goto cont
b2:
  mut b = 1
  a <- b
  print a
cont:
```

Conclusion: moving around code can invalidate your proposed invariant. Recovering it would be complicated.

Scopes are easier to preserve since they have only one starting point (ie. the declaration). But the undefined/defined can have arbitrarily many assignments to the same variable.

Therefore I'd like to lift this requirement for now.

o-:
superseeded by #60

gasche:
(Thinking out loud, this is not in contradiction to what you say.) If `a` was used after `cont` in your example, then the proposed rewriting would be wrong: you would need to assign `a <- a0` in `b1`. If `a` is not used, and `clear,drop` were maximally placed, then the issue would not occur: before both transitions to `cont` we would have a `drop` if `a` is not mentioned anymore, and a `clear` if `a` is assigned but not required.

o-:
> If a was used after cont in your example, then the proposed rewriting would be wrong:

agreed. In the parlance of #58 the assignment would not dominate all its uses (therefore no lifting will be done).

> If a is not used, and clear,drop were maximally placed, then the issue would not occur:

I think you might be right. Let's keep this in mind! when we have resolved the issue of how to place drops and clears maybe we can resurrect the invariant.

PR #60, opened 2017-02-21
Scope uninitialized rebased

o-:
the same as #59 but with clean history


o-:
@gasche are you ok with this for now?

gasche:
I'm ok with the change -- and the implementation. Let's give @mhyee and @ulysses4ever a chance to ask questions about the implementation if they have some, then merge.

mhyee:
Looks good to me.

PR #62, opened 2017-02-28
transformation to minimize lifetimes of variables

o-:
this transformation removes dead stores, dead declarations and places
drops to minimize the lifetime of variables according to the scoping
rules.

### DIFF
--- a/analysis.ml
+++ b/analysis.ml
@@ -186,17 +186,22 @@ let scope_analysis (introduction : instruction -> variable list)
   let update pc cur_scope =
     let instr = instrs.(pc) in
     let to_remove = introduction instr in
-    let remove acc var = VariableMap.remove var acc in
-    let cur_scope' = List.fold_left remove cur_scope to_remove in
+    let cur_scope = List.fold_right VariableMap.remove to_remove cur_scope in
     let to_add = elimination instr in
-    let merge acc var = VariableMap.union (VariableMap.singleton var pc) acc in
-    List.fold_left merge cur_scope' to_add
+    let entry var = VariableMap.singleton var pc in
+    let merge acc var = VariableMap.union (entry var) acc in
+    List.fold_left merge cur_scope to_add
   in
   backwards_analysis VariableMap.empty instrs merge update
 
 let liveness_analysis (instrs : segment) =
   let introduction instr = VarSet.elements (ModedVarSet.untyped (defined_vars instr)) in
   let elimination instr = VarSet.elements (used_vars instr) in
+  scope_analysis introduction elimination instrs
+
+let lifetime_analysis (instrs : segment) =
+  let introduction instr = VarSet.elements (ModedVarSet.untyped (declared_vars instr)) in
+  let elimination instr = VarSet.elements (required_vars instr) in
   scope_analysis introduction elimination instrs
 
 (* returns a 'pc -> variable set' computing live vars at a certain pc *)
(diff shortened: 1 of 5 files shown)


analysis.ml
```diff
@@ -177,26 +177,27 @@ let reaching (instrs : segment) : pc -> PcSet.t =
...
-    let kill = VarSet.elements (ModedVarSet.untyped (defined_vars instr)) in
+    let to_remove = introduction instr in
     let remove acc var = VariableMap.remove var acc in
-    let uses = List.fold_left remove uses kill in
-    (* Then add used vars *)
-    let used = VarSet.elements (used_vars instr) in
-    let merge acc var = VariableMap.union (VariableMap.singleton var pc) acc
-    in
-    List.fold_left merge uses used
+    let cur_scope' = List.fold_left remove cur_scope to_remove in
```
gasche: you could use `List.fold_right VariableMap.remove to_remove cur_scope` here.

analysis.ml
```diff
@@ -177,26 +177,27 @@ let reaching (instrs : segment) : pc -> PcSet.t =
...
+    let to_add = elimination instr in
+    let merge acc var = VariableMap.union (VariableMap.singleton var pc) acc in
+    List.fold_left merge cur_scope' to_add
```
gasche: `List.fold_right (fun v -> VariableMap.add v pc) to_add cur_scope` works here (it's a bit clever, but idiomatic)

analysis.ml
```diff
@@ -228,3 +229,38 @@ let dominates (instrs : segment) : pc -> pc -> bool =
   fun pc pc' ->
     let doms = dominators pc' in
     PcSet.exists ((=) pc) doms
+
+let lifetime_analysis (instrs : segment) =
+  let introduction instr = VarSet.elements (ModedVarSet.untyped (declared_vars instr)) in
+  let elimination instr = VarSet.elements (required_vars instr) in
+  scope_analysis introduction elimination instrs
```
gasche: It would make sense to move that definition just below `liveness_analysis`: they are related and this one does not anything declared in between.

analysis.ml
```diff
@@ -228,3 +229,38 @@ let dominates (instrs : segment) : pc -> pc -> bool =
...
+
+(* returns a 'pc -> pc set' computing the set of instructions depending on a declaration *)
+let required (instrs : segment) : pc -> PcSet.t =
+  let res = lifetime_analysis instrs in
+  fun pc ->
+    let instr = instrs.(pc) in
+    let declared = VarSet.elements (ModedVarSet.untyped (declared_vars instr)) in
+    let required_of var = VariableMap.at var (res pc) in
+    let all_requires = List.map required_of declared in
+    List.fold_left PcSet.union PcSet.empty all_requires
```
gasche: This works and is rather clear. If you wanted, you could easily avoid the conversion to lists and back:

```ocaml
    let add_uses (_, x) used = PcSet.union used (VariableMap.at x (res pc)) in
    ModedVarSet.fold add_uses (defined_vars instrs.(pc)) PcSet.empty
```

analysis.ml
```diff
@@ -228,3 +229,38 @@ let dominates (instrs : segment) : pc -> pc -> bool =
...
+(* The same as required_vars_at but extends the required interval to
+ * merge points to conform to our scoping rules *)
+let required_merged_vars_at instrs =
+  let required_vars_at = required_vars_at instrs in
+  let merge pc cur_lifetime in_lifetime =
+    let merged = VarSet.union cur_lifetime in_lifetime in
+    if VarSet.equal cur_lifetime merged then None else Some merged
+  in
+  let update pc cur_lifetime = VarSet.of_list (required_vars_at pc) in
+  forward_analysis VarSet.empty instrs merge update
```
gasche: I think we could define a generic `saturate_merge` operation that takes the result of a data-flow analysis and uses a forward_analysis path to merge joining paths. Then this would be `saturated_merge (required_vars_at instrs)`.

gasche (approved):
I think the code is ok, I did some style comments.

Given that `required_vars_at` is not used outside its defining module while `required_merged_vars_at` is used, I think that a final version of the PR should give the convenient name to the actually-exported function: what is today `required_merged_vars_at` should be named `required_vars_at`.

mhyee:
The code also looks good to me.

analysis.ml
o-: that does not work! We want {x -> 2} and {x -> 3} to merge to {x -> {2,3}}

PR #64, opened 2017-03-06
[WIP] add functions

o-:
This is playing fast and loose. It probably does not work. But I
would like to start a discussion!

What do you think?

### DIFF
--- a/doc/ir-semantics.text
+++ b/doc/ir-semantics.text
@@ -7,8 +7,8 @@ F ::= Segment  active ↦ S, (Ds ↦ S)*   function:           active version an
 S ::= start ↦ i,           I           segment:            instruction stream with dedicated start
 I ::= (l ↦ i)*                         instruction stream: labeled instructions
 
-Df ::= Function f(x, ...)   function declaration
-Ds ::= Segment  s           segment declaration
+Df ::= Function f[x, ...](y, ...)  function declaration
+Ds ::= Segment  s                  segment declaration
 
 f  function name
 s  optimization version
@@ -40,7 +40,7 @@ e ::=    (simple) expressions
 | lit                   literals
 | x                     variables
 | primop(x, ...)        primitive operation (pure)
-| close f               create closure
+| close f[x, ...]       create closure
 
 lit ::=  litterals
 | nil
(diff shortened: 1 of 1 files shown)


o-:
How should we implement free variables in closures?

The easiest would be to have a "lookup x" expression which looks up x in the current environment. The difference between `const a = x` and `const a = lookup x` would be that the latter is valid even if x is not in scope.

This gives us few guarantees. So maybe a function declaration should include a list of free variables and `close` will be guaranteed to create a closure environment with exactly those.

gasche:
> How should we implement free variables in closures?

I don't know what you mean. In the lambda-calculus, we don't do anything special for free variables: a variable refers to its nearest definition in scope -- all variables have to be "in scope", but we don't use a different way to address variables that are parameters of the nearest enclosing function definition and the others.

gasche:
After thinking harder about the proposed design, now I know what you mean :-)

One design that we *could* have chosen is to write function definitions at any point in the middle of the instruction stream, to define a local function/closure.

```
const x = 1
def f{x}(y):  // captures `x`, takes `y` as parameter
  return (x + y)
const z = call f (2)
print z
```

One problem with that is that it makes the instruction stream more structured, with a tree-like structure (are the `def()` and all the following code a single instruction like, say, stack languages suspended code blocks are just one item on the stack?).

What you propose instead is to have functions only at the toplevel, which allows to keep a less-structure instruction stream (no function declarations at arbitrary points in the bytecode). I would suggest to write the previous example in the following way

```
def f{x}(y):
  return (x + y)

def main():
  const x = 1
  const x' = closure f{x}
  const z = call x'(2)
  print z
```

In other terms, our functions declarations are lifted at the toplevel, but they make an explicit distinction between environment-passed variables and parameter-passed variables. At the closure-creation site you explicitly pass the captured environment. This is extremely similar to closure constructions in the OCaml bytecode language for example (but variable-based instead of stack-based).

(Higher-level languages do not distinguish the place of declaration of a (local) function from the place where the captured environment is determined. Making the distinction allows for a bit more flexibility  (we can build a closure in slightly different ways in two branches, instead of having to define all its variables differently and define the closure after the merge point), I'm not sure we actually need/want the flexibility, but if that's closure to the look&feel of functions/closures in other bytecode languages that's probably better.)

o-:
yes, sorry, some more explanations on my side would have helped.

I favor the design of keeping functions at the top-level because it makes the structure more regular. For example there is no inherent distinction between functions and closures. Also as you mention the inline def-block creates some representation challenges in the bytecode format.

I like your syntax proposal for the declarations and will try to integrate it into the design doc. If we make the arguments optional we can have the convenience and the regularity:

```
Function main:
  ...

Function top-level(args):
  ...

Function closure{capture}(args):
  ...
```

UPDATE: I saw you call the expression ```closure```. Do you mind if we call it ```close``` as it takes a function and closes it over an environment to create a closure :)

gasche:
I'm not fond of the "no inherent distinction between (closed) functions and closures" idea. They have a different runtime representation (a function pointer vs. a pair) and the call has a different runtime semantics. I would like one of the following more:

- we can choose to hide closed functions by making `def f(args): ...` syntatic sugar for `def f[](args): ...` followed by `const f = closure f[]`. Then all function identifiers we see are always closures (some with an empty environment). Is this what you have in mind?
- or we could have an explicit distinction with `def f(args): ...` and a dedicated `direct_call x_f (vals)` construction.

gasche:
Yes, I'm fine with `close`, sorry.

o-:
Aha, I see what you mean.

My idea was to only have closures (no naked function pointers at all). In this scheme the interpreter starts `main` in an empty environment. But every subsequent `call` has to call a closure closed over some environment. Of course a compiler would have to prologue `main` with a sequence to initialize all library functions (eg. `const add = close add[]`).

Or like you suggest we could say we have a global environment (ie. the environment associated with main) and every top-level function with empty capture is implicitly bound in the main environment. But note: currently function lookup is the same as variable lookup. This means even in this scheme every function we want to call from a closure needs to be explicitly captured by that closure. I am not convinced that this is a good property...

gasche:
So your idea was to have `def f(args)` as syntactic sugar for `def f{}(args)`, but keep the `close f{}` part explicit. If we are going to be explicit, why not also write `def f{}` explicitly? (I would propose to have just `def main:`, or any other syntax for main that is *distinct* from a function declaration, as this is a distinguished object).

(I'm not pushing in favor of having separated function pointers in the language; in a realistic language that's an important optimization to make, but here I am not sure that it adds much for our purposes, so I'm fine with keeping them out.)

(Apologies for going back and forth between `[]` and `{}` for closure environments. I started with `[]`, but then github started turning `f[x](args)` into a link (within a code block...), so I moved to `{}`. I'm fine with either.)

o-:
superseded by #67

PR #65, opened 2017-03-12
Small steps

gasche:
This was fun.

### DIFF
--- a/rewrite.ml
+++ b/rewrite.ml
@@ -36,40 +36,20 @@ type push_status =
   | Need_pull of pc
   | Work of Edit.result * pc list
 
-let push_drop instrs dropped_var pc_drop =
-  let pc_above = pc_drop - 1 in
-  assert (instrs.(pc_drop) = Drop dropped_var);
+let push_instr is_eliminating is_anihilating is_blocking instrs pc : push_status =
+  assert (pc >0);
+  let pc_above = pc - 1 in
+  let to_move = instrs.(pc) in
   let instr = instrs.(pc_above) in
-  let _succs = Analysis.successors instrs in
   let preds = Analysis.predecessors instrs in
-  let blocking_instr instr =
-    VarSet.mem dropped_var (required_vars instr) in
   let edit new_instrs =
     Edit.subst instrs pc_above 2 new_instrs in
-  let general_regime instr =
-    if blocking_instr instr then Blocked
-    else Work (edit [|Drop dropped_var; instr|], [pc_above]) in
-  match instr with
-  | Assign (x, _) | Clear x ->
-    (* if the current instruction defined the dropped variable,
-       we can delete the instruction *)
-    if x = dropped_var
-    then Work (edit [|Drop dropped_var|], [pc_above])
-    else general_regime instr
-  | Decl_const (x, _) | Decl_mut (x, _) ->
-    (* if the current instruction declared the dropped variable, we
-       can delete the instruction, and we must also forget about the
-       drop *)
-    if x = dropped_var
-    then Stop (edit [||])
-    else general_regime instr
+  match[@warning "-4"] instr with
   | Stop ->
     (* we are coming from the predecessors, cannot reach a Stop! *)
     assert false
-  | (Read _ | Branch _ | Osr _ | Drop _ | Goto _ | Comment _ | Print _) as instr ->
-    general_regime instr
   | Label label
-    when List.exists (fun pc -> blocking_instr instrs.(pc)) preds.(pc_above)
+    when List.exists (fun pc -> is_blocking instrs.(pc)) preds.(pc_above)
     -> Blocked
   | Label label ->
     (*
@@ -95,7 +75,7 @@ let push_drop instrs dropped_var pc_drop =
           branch edges from our (multi)predecessors, so that the
           Not_found case below will do the final push. *)
        let (_instrs, pc_map) as edit = split_edge instrs branch_pc label pc_above in
-       Work (edit, [pc_map pc_drop])
+       Work (edit, [pc_map pc])
    | exception Not_found ->
      (* multi-predecessor case, with no multi-successor predecessor (no branch);
         we can move the drop above all predecessors. For a predecessor in (pc),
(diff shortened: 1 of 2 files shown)


rewrite.ml
```diff
@@ -0,0 +1,182 @@
+open Instr
+
+(* TODO remove the fallthrough cases *)
```
o-: we did that

o-: @gasche you have to fix it, the PR is on your fork :)

gasche: Sorry, fixed in fdf6cc61481d017cdc6502fc0e540a9e6ec28c12.

edit.ml
```diff
@@ -0,0 +1,46 @@
...
+    end
+  in
+  move from_pc;
+  instrs.(to_pc) <- from
+
+type pc_map = int -> int
+
+type result = Instr.instruction_stream * pc_map
+
+let subst_pc instrs start_pc len new_instrs pc =
```
mhyee: Could you add some documentation for this (and also `subst` and `subst_many`)? For example, what the parameters mean, and what the function is actually used for.

For `subst`, it didn't occur to me that `len` could be `0`, until I saw it being used that way. Probably because I thought `len` referred to the length of `new_instrs`.

For `subst_many`, documenting the `substs` parameter would have been helpful. I also object to reusing the variable name `rest` for completely different things (i.e. the rest of `substs` and the rest of the result).

instr.ml
```diff
@@ -261,3 +261,8 @@ module Value = struct
   let int n : value = Lit (Int n)
   let bool b : value = Lit (Bool b)
 end
+
+let instr_at instrs pc =
+  if Array.length instrs = pc then Stop
```
mhyee: What if `Array.length instrs > pc`? Did I miss something?

rewrite.ml
```diff
@@ -0,0 +1,196 @@
+open Instr
+
+(** [split_edge instrs pc label pc']
```
mhyee: This documentation is good, but it's still not obvious what "splitting the edge" means. I think a short statement like "The result is that [pc] is no longer a predecessor of [pc']." Or maybe even a short example.

rewrite.ml
```diff
@@ -0,0 +1,196 @@
...
+    let insert = (pc_branch, 0, [| drop_instr |]) in
+    let delete label_pc = (label_pc + 1, 1, [||]) in
+    let deletes = List.map delete succs.(pc_branch) in
+    let (_instrs, pc_map) as edit = Edit.subst_many instrs (insert :: deletes) in
+    Pulled_to (edit, pc_map pc_branch - 1)
+  end
+
+type progress = Made_progress | Stuck
+
+let debug =false
```
mhyee: Is this still used?

tests.ml
```diff
@@ -571,6 +570,290 @@ let do_test_codemotion = function () ->
...
 
+let do_test_pull_drop () =
+  let open Rewrite in
+  let main = List.assoc "main" in
+  let test input pc x expected =
+    let input = main input in
+    let instrs = match Rewrite.pull_drop input x pc with
+      | Pulled_to ((instrs, _), _) -> instrs
+      | Blocked -> input in
+    (* Printf.printf "%s:\n%s\n"
```
mhyee: Is this printing for debugging? Or should it just be removed (instead of being commented out)?

There's another instance of this below, for the push drop tests.

tests.ml
```diff
@@ -663,6 +946,8 @@ let suite =
    "used">:: do_test_used;
    "liveness">:: do_test_liveness;
    "codemotion">:: do_test_codemotion;
+   (* "push_drop">:: do_test_push_drop; *)
```
mhyee: Why is this commented out?

mhyee (commented):
I'm going to be that person who asks for more documentation. I've commented on some of the places I had trouble following.

More generally, I think the whole `push_drop` and `pull_drop` needs some sort of description or summary. Same with `move_drops` and `move_drops_on_var`. The email from a while back was good, but it's something I had to dig up when reviewing this code. For example, I liked the summary:

>   - if the instruction depends on `x`, you cannot move above
>  
>   - if the instruction does not use `x` at all, you can move
>     above
> 
>   - if the instruction assigns `x` (`x <- e`), you can move
>     above *and* delete the instruction (optimization!);
>     (for `read` we would need an anonymous read `read _` to
>     leave in place)
> 
>   - if the instruction declares `x`, you can delete it *and*
>     delete the drop as well.

Explaining the need for edge splitting (and defining what it exactly is) would be good.

Also, just from reading the names "push" and "pull," it's not clear what "direction" they're referring to. I can't think of any short names that would make this clearer, so this is just another request for documentation.

Finally, it seems there's a few merge conflicts. Maybe there's some GitHub issue -- it looked like the commits were out of order. I'm not sure what happened.

gasche:
These are excellent comments, thanks. I'll try to iterate to fix those as soon as possible.

o-:
@gasche @mhyee  
I am currently working on a PR that abstracts over the drop moving to have a generic move-instruction version. If you all agree I would suggest that we merge this PR and I can address the review in the new PR? Because otherwise it might become a bit messy.

gasche:
I will do one round of iteration tomorrow morning to fix the conflicts and merge the branch as you suggest. I may fix some of the issues reported at the same time.

mhyee:
Yeah, that sounds good.

o-:
@gasche  
oh, ok. I was too fast on this one, sorry.
https://github.com/reactorlabs/sourir/commit/df7ddc33d0896053ce5f072494ad217710190c05
What shall we do? If you want to clean up this one first you can go ahead. I'll figure out how to rebase. Otherwise you can cherry-pick that commit and continue on top in this PR?

gasche:
I pushed your commit on top of the PR. If you make other commits to your
`move-it` branch I can push them as well.

P.S.: We are starting to have a bit too many branch in our forks; I did a
round of `git branch -d ...`, and then `git push --all --prune gasche`.

o-:
@gasche 
thank you! I might start playing with moving around other instructions. For example assignments or osr.

rewrite.ml
```diff
@@ -0,0 +1,228 @@
...
+  new_instrs, pc_map
+
+type push_status =
+  | Stop of Edit.result
+  | Blocked
+  | Need_pull of pc
+  | Work of Edit.result * pc list
+
+let push_instr is_eliminating is_anihilating is_blocking instrs pc : push_status =
+  assert (pc >0);
```
mhyee: Nitpick: missing space.

rewrite.ml
```diff
@@ -0,0 +1,228 @@
...
+      | Made_progress ->
+        Some instrs
+  in
+  work_push instrs Stuck pc []
+
+let move_up is_target is_eliminating is_anihilating is_blocking instrs =
+    let move_up_instr = move_up_instr is_eliminating is_anihilating is_blocking in
+
+    (* Loop around and move all drops until none moves anymore *)
+    let rec work status instrs pc =
```
mhyee: This function is a bit confusing, because `instrs` is reused. It's particularly bad in lines 171/172 where both cases return `instrs` but they're different. 177/178 is similar.

mhyee: It's actually worse because there are *three* `instrs`. One is an argument, one is the let binding on line 170, and one is from `Some instrs`.

rewrite.ml
```diff
@@ -0,0 +1,228 @@
...
+      else if is_target instrs.(pc) then
+        begin match move_up_instr instrs [pc] with
+          | None -> work status instrs (pc+1)
+          | Some instrs -> work Made_progress instrs 0
+        end
+      else
+        work status instrs (pc+1)
+    in
+    work Stuck instrs 0
+
```
mhyee, line 187: Delete this empty line?

rewrite.ml
```diff
@@ -0,0 +1,228 @@
...
+
+module Drop = struct
+  let is_blocking var instr =
```
mhyee: Note that in the arguments of `move_up` and on lines 94-96, the order is `is_eliminating`, `is_annihilating`, (it's spelled with two **n**s) and `is_blocking`. But here, the order is `is_blocking`, `is_eliminating`, and `is_annihilating`.

rewrite.ml
```diff
@@ -0,0 +1,228 @@
...
+    | Decl_const (x, _) | Decl_mut (x, _) -> x = var
+    | _ -> false
+
+  let move_up_var instrs var =
+    let is_target instr =
+      match[@warning "-4"] instr with
+      | Drop x when x = var -> true
+      | _ -> false
+    in
+    let move = move_up
```
mhyee: It's a bit confusing to have two different functions called `move_up`. Especially when it's **not** the function defined just a few lines later.

mhyee:
I added some more comments on code style. Other than that, it looks fine.

Generalizing the algorithm for other instructions is good, but I wonder if it's general enough (but not too general). It'll probably be fine, and we can tweak it later. I'm basically citing [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) and the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)), but since we do want to move other instructions, I think the refactor is justified.

o-:
@mhyee thanks a lot for looking at it. My patch was a bit sloppy as I was rushing to get it done to not block progress. It does not even address your earlier comments. Don't waste too much time on it. I am not sure yet if it will be general enough/too general. But I think we'll see. At least the moves I am playing around with right now seem to mostly use is_blocking. especially is_annihilating is very particular to drop. But the whole machinery to drive the moves is certainly a general thing. Also there seems to be some bugs regarding termination. At least my osr moving loops forever. I guess I'll have another patch soon :)

gasche:
I just rebased, and plan to merge.

gasche:
So I merged the first part (the part Oli and I did together, for which @mhyee's comments had already been handled), and I applied some additional patches suggested by @mhyee's comments on the last commit by @o-. Any final comments?

o-:
lgtm, thanks for the additional cleanup.

PR #66, opened 2017-03-15
Define step relation by structural induction

o-:
For #67 I reorganized the reduction relation a bit. I would like to keep the work from there.

Reorganize the reduction relation to take the currently being
evaluated instruction as an additional argument. This allows us
to (i) describe the relation by structural induction on the
instruction and (ii) makes it a lot easier to talk about
instructions which add/remove to the continuations stack.

### DIFF
--- a/doc/ir-semantics.text
+++ b/doc/ir-semantics.text
@@ -2,35 +2,37 @@
 
 x        variables
 
-l        labels
-v        optimization version
+P ::= Segment  active   ↦ S, (Ds ↦ S)*   program:            active version and others
+S ::= start ↦ i,          I              segment:            instruction stream with dedicated start
+I ::= (l ↦ i)*                           instruction stream: labeled instructions
 
-s ∈ l    start label
+Ds ::= Segment  s                  segment declaration
 
-a        addresses
+s  optimization version
+l  labels
 
-P ::= (v ↦ M)*             program
-M ::= (s ↦ i), I
-      s ∉ I                module
-I ::= (l ↦ i)*             instruction stream
+active    active segment
+start     start label
+
+a  addresses
 
 i ::=    instructions
-| const x = e               constant variable
-| mut x = e                 mutable variable
-| x <- e                    assignment
-| branch x l₁ l₂            conditional
-| goto l                    goto
-| print x                   print
-| osr(e, v, l, (x = ee)*)   osr : (can be used for osr-in and osr-out)
-
-ee ::=   extended expression
-| e
-| ⊥
+| const x = c                   constant variable
+| mut x = c                     mutable variable
+| x <- c                        assignment
+| branch e l₁ l₂                conditional
+| goto l                        goto
+| print e                       print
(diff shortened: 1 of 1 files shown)


o-:
Btw. this also addresses the divergence to the interpreter. In this updated version of the language instructions like `print` take expressions as arguments and not just identifiers.

doc/ir-semantics.text
```diff
@@ -53,63 +75,77 @@ succ I l =
...
+  step   (const x = v)
+         (P, T, H, E* E, C*)
+       = (P, T, H, E* (E, x ↦ v), stepi C*)
 
-  (T, H, E, P, I, l) -> (T, H, (E, x ↦ v), P, I, succ I l)
-    when I(l) = (const x = e)
-     and v := eval H E x
+  step   (mut x = v)
+         (P, T, H, E* E, C*)
+       = (P, T, H[a ↦ v], E* E[x ↦ v], stepi C*)
```
mhyee: Should be `E[x ↦ a]`

doc/ir-semantics.text
```diff
@@ -53,63 +75,77 @@ succ I l =
...
-     and v := eval H E x
+  step   osr(v, _, _)
+         (P, T, H, E* E, C*)
+       = (P, T, H, E* E, stepi C*)
+    when v = false
 
-  (T, H, E, P, I, l) -> (T, H, E, P, I, succ I l)
-    when I(l) = osr(e, _, _)
-     and (eval H E e) is not (true)
+  step   osr(e, s', l', (x = ee, ...))
```
mhyee: Should be `osr(v, s', l', ...)`

doc/ir-semantics.text
```diff
@@ -53,63 +75,77 @@ succ I l =
...
+       = (P, T, H[a ↦ v], E* E[x ↦ v], stepi C*)
+    when a fresh
 
-  (T, H, E, P, I, l) -> (T, (H, a ↦ v), (E, x ↦ a), P, I, succ I l)
-    when I(l) = (mut x = e)
-     and v := eval H E e
-     and a fresh
+  step   (x ← v)
+         (P, T, H, E* E, C*)
+       = (P, T, (H,E)[x] := v, E* E, stepi C*)
```
mhyee: Looking at the definition on [line 87](https://github.com/reactorlabs/sourir/pull/66/files/fe69288fc7ff17bc9f82ed821e9c874237c2ac55#diff-47b6a92f844b6cdae62b27728d90d5eeR86), I guess the notation should be `(H,E)[x ← v]`

mhyee:
I found two typos and one instance of inconsistent notation. Overall, I find it a lot more readable than the previous version!

o-:
thanks for spotting the typos.

gasche:
(This new style should also make it easier to have function calls on top; previously we had a separate variable binding for function calls because we did not want to define instruction-level contexts.)

PR #68, opened 2017-03-19
optimized branch pruning

o-:
This is a revised implementation of branch pruning. Instead of just replacing `guard` by `osr` it will try to push the `osr` instruction up the call graph.

As long as the intermediate instructions do not change the value of the expression `exp` in `Branch (exp, l1, l2)` the guard can safely be performed earlier.

If while moving up the osr we encounter a pc where we have moved across every predecessor before, we can drop it. The assumption is that at least one of the predecessors must lead to a dominating pc
and that's where the final deopt will be placed. The only case where this goes wrong is when the entry instruction does not dominate us (ie. "loop: goto loop"). To rule out this case we add a comment
instruction as the first instruction (which is guaranteed to dominate all others).

Btw. this optimization introduces a new implementation technique where instead of using a pc_mapping we instrument the original code with a label at every instruction. This allows us to find back to the original position in unoptimized code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,11 +30,11 @@ test_examples: sourir
 	mkdir $(TEMPDIR)/examples
 	for f in examples/*.sou; do yes 0 | ./sourir $$f --quiet \
 	  > $(TEMPDIR)/$$f.out; done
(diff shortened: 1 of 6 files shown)


o-:
I put this on hold for a bit. I have a revised version in mind that comes with a proof sketch and is much more universal.

PR #70, opened 2017-03-21
add top-level static functions, with static call and return

o-:


doc/ir-semantics.text
```diff
@@ -108,18 +118,30 @@ Update (partial) function, returns a S:
        = step CC[v] (P, T, H, E*, C*)
     when v := eval P H E e
 
+  step   CC[call f (v, ...)]
+         (P, T, H, E*, C*)
+       = (P, T, H, E* E', C* C)
```
gasche: Can't we store the current lexical environment away (to reinstate it on `return`)?

Also `C* C`, you mean `C* CC`? (Or conversely `CC[call ..]` -> `C[call ...])

doc/ir-semantics.text
```diff
@@ -2,34 +2,42 @@
...
-| const x = c                   constant variable
-| mut x = c                     mutable variable
-| x <- c                        assignment
+| const x = ce                  constant variable
+| mut x = ce                    mutable variable
+| x <- ce                       assignment
 | branch e l₁ l₂                conditional
 | goto l                        goto
 | print e                       print
+| return e                      return
```
gasche: So why is `print ce` not allowed? It would be simpler to not make the distinction, but I'm sure there is a rationale -- I thought it would be easier when `print [.]` is already a continuation.

doc/ir-semantics.text
```diff
@@ -108,18 +118,30 @@ Update (partial) function, returns a S:
...
+         (P, T, H, E*, C*)
+       = (P, T, H, E* E', C* C)
+    when P(f)  = Function (x, ...)
+     and E'   := (x ↦ v), ...
+     and C    := (f, active, start)
+
+  step   (return v)
+         (P, T, H, E* E', C* C C')
+       = step cs (P, T, H, E*, C* C)
+    when cs := P(C)[call(...)/v]
```
gasche: urgh.

doc/ir-semantics.text
o-: Ok I see this is causing some confusion.

CC is just an execution context, the only thing it really helps us is in applying eval to an instruction in the right order.

C is the continuation and it contains just the label, not the actual expression. but I guess you are right that by storing the expression in the continuation it will be a lot more intuitive.

o-: Oh and why is the call restricted to the assignments, there is no good rationale.

I certainly do not want the call to be a normal expression for now, because then expressions are not pure anymore (ie. no `call() + call()`).

In the case of `print e` it could be `ce` as well. I thought it might be easier to reason if you cannot do a side-effect and a call in one instruction.

doc/ir-semantics.text
o-: see above, the whole C and CC business is causing confusion. `CC` is really only used to pattern match in `step`.

doc/ir-semantics.text
o-: okok, I am just saying, that so far it was possible to have no expressions in continuations. If you want me to fix this, then we will have expressions in continuations.

doc/ir-semantics.text
o-: Maybe to give you an example of how I want this to work:

for

```
main:
  1: const y = call f (42)
  2: stop

Function f(x):
  1: return x
```

consider `step instruction continuation env` then

```
  step  (CC[call f (1)])  (main, 1) []
=>
  step (return x) ((f, 1), (main, 1)) [x -> 42]
=>
  step (const y = 42) (main, 1) []
=>
  step (stop) (main, 2) [y -> 42]
```

doc/ir-semantics.text
```diff
@@ -2,34 +2,42 @@
 
 x        variables
 
-P ::= Segment  active   ↦ S, (Ds ↦ S)*   program:            active version and others
+P ::= Function main    ↦ F, (Df ↦ F)*    program:            main function and other functions
+F ::= Segment  active  ↦ S, (Ds ↦ S)*    function:           active version and others
 S ::= start ↦ i,          I              segment:            instruction stream with dedicated start
 I ::= (l ↦ i)*                           instruction stream: labeled instructions
 
 Ds ::= Segment  s                  segment declaration
```
mhyee: Missing the definition for `Df`, which was used above in line 5.

doc/ir-semantics.text
```diff
@@ -100,49 +110,62 @@ Update (partial) function, returns a S:
...
+       = (P, T, H[a ↦ v], E* E[x ↦ a], stepi π*)
     when a fresh
 
   step   (x ← v)
-         (P, T, H, E* E, C*)
-       = (P, T, (H,E)[x ← v], E* E, stepi C*)
+         (P, T, H, E* E, π*)
+       = (P, T, (H,E)[x ← v], E* E, stepi π*)
 
   step   osr(v, _, _)
```
mhyee: Typo, should be `osr(v, _, _, _)`.

doc/ir-semantics.text
```diff
@@ -100,49 +110,62 @@ Update (partial) function, returns a S:
...
-         (P, T, H, E* E, C*)
-       = (P, T, H, E* E, stepi C*)
+         (P, T, H, E* E, π*)
+       = (P, T, H, E* E, stepi π*)
     when v = false
 
   step   osr(e, v', l', (x = ee, ...))
```
mhyee: This should be:
```
step   osr(v, s', l', (x = ee, ...))
       (P, T, H, E* E, π* (f, s, l))
     = (P, T, H, E* E', π* (f, s', l))
```

doc/ir-semantics.text
```diff
@@ -100,49 +111,62 @@ Update (partial) function, returns a S:
...
+         (P, T, H, E*, π*)
+       = step CC[v] (P, T, H, E*, π*)
     when v := eval P H E e
 
+  step   CC[call f (v, ...)]
+         (P, T, H, E*, π*)
+       = (P, T, H, E* E', π* π)
+    when P(f)  = Function (x, ...)
+     and E'   := (x ↦ v), ...
+     and π    := (f, active, start)
```
mhyee: Is the segment and label always going to be `active, start`?

doc/ir-semantics.text
o-: What's the diff you want? It should really be `l'` and not `l`. The osr defines both a `v` and a `l` to continue

mhyee: Yeah, sorry, I rushed and also had a typo.

`e` should be `v` and `v` should be `s`. And then `v'` should be `s'`.

doc/ir-semantics.text
o-: the idea is that by definition active is the most up-to-date version of the function and start is the first instruction

o-:
@mhyee : ok finally I think I have fixed all f*ups. thanks again
@gasche : given this version, what do you prefer? The expression retrieval hack at https://github.com/reactorlabs/sourir/pull/70/files#diff-47b6a92f844b6cdae62b27728d90d5eeR137 or proper continuations?

gasche:
The `call ...` substitution looks like a design smell to me, but I don't have the brain cycles right now to think about the continuation route and ensure that it is better. Having full continuations would be mildly irritating as we feel that we restricted ourselves to a linear language and ANF form to avoid some of that conceptual cost, but this may still be the cleaner choice -- and I'm not even sure that the intuition on what the low-level-ness of bytecode should give us or prevent us from having is right.

(One thing that I definitely feel we should be able to avoid is concatenation of the new environment to the old environment on function entry. We should be able to store the old environment on the side and not have it available at all within the function. Implementation wise the current design smells a bit.)

o-:
@gasche : Thanks for those excellent comments.

I have quite reworked the proposal (and again renamed segments since `M` was already taken for the machine state -- i feel stupid)

1. Regarding the call substitution I am now gravitating towards the initial `const x = call ...` construct. The reasoning is actually quite simpre: an interpreter would store the call result either in an accumulator register or on the stack. That's why most interpreters can implement `call` as an instruction -- the storing of the result is implicit. The closest thing we have to an accumulator register is a const variable. Therefore I believe this should be the only possible way of invoking a function if you care about the result.
2. Regarding the program having access to the environment stack, that was certainly not my intention. I tried to reorder the whole thing. Now the machine state has two parts separated by a `:`. The first part which holds the continuation is unreachable by the intra function `step` reduction.

>  and I'm not even sure that the intuition on what the low-level-ness of bytecode should give us or prevent us from having is right

ok, here I have to disagree. The reason was to be able to easily argue that the model applies to an actual interpreter. That's why I am trying really hard to not have unrealistic things (like a stack of expressions) in there.

doc/ir-semantics.text
```diff
@@ -102,77 +90,71 @@ Update (partial) function, returns an H:
...
+  step   (const x = e) (P, I, T, H, E, l)
+       = (I, T, H, E[x ↦ v], succ I l)
+    when v := eval H E e
+
+  step   (mut x = e) (P, I, T, H, E, l)
+       = (I, T, H[a ↦ v], E[x ↦ a], succ I l)
+    when v := eval H E e
+     and a - fresh
+
+  step   (x ← v) (P, I, T, H, E, l)
```
mhyee: Typo, should be: `step   (x ← e) (P, I, T, H, E, l)`

o-:
If no further comments until monday I'll merge it

gasche:
Either this or #71 are fine by me.

PR #72, opened 2017-03-21
disambiguate segment and scope meta-variable

o-:


doc/ir-semantics.text
```diff
@@ -2,8 +2,8 @@
 
 x        variables
 
-P ::= Segment  active   ↦ S, (Ds ↦ S)*   program:            active version and others
-S ::= start ↦ i,          I              segment:            instruction stream with dedicated start
+P ::= Segment  active   ↦ M, (Ds ↦ M)*   program:            active version and others
```
mhyee, line 5: Do you still want to keep `Ds` and little `s`? (And where does `M` come from?)

o-: M is defined bellow.

mhyee: Sorry, I meant why `M` was chosen. The definition below is fine.

o-: ah segMent :/ do you have a better idea?

mhyee: I don't have a better idea, but I think it's reasonable. I was just curious.

o-:
thanks for the comments.

I am really sorry for making a mess. I'll push this commit to #70 instead, then we can fix all of it in one go.

doc/ir-semantics.text
o-: btw. I use `T` for segmenT now, since it avoids confusion with Module and also I found Dm to look silly.

Issue #73, opened 2017-03-22
Constant propagation/folding


mhyee:
I've been pair programming this with @o-.

mhyee:
I've decided to punt on copy propagation. The naive implementation of copy propagation won't work:

Original program:
```
mut x = 0
const y = x
x <- 1
print y   # Outputs 0
```

After constant propagation:
```
mut x = 0
const y = x
x <- 1
print x   # Outputs 1
```

Of course, I could check that `x` is a constant variable before propagating it. Maybe this is simple to implement, but I think focusing on constant folding will give us more interesting examples. I can always revisit this if we decide it's something we want.

mhyee:
Constant folding and constant propagation rely on each other:

```
const x = 1
const y = (2 + x)  # Requires constant propagation of x
const z = (3 + y)  # Requires constant folding of y, then constant propagation of y
```

To fully optimize the above code, you need to perform propagation, then folding, then propagation again, then folding again.

My original plan was to keep the constant propagation and constant folding as small and separate optimizations, and then iterate as long as some optimization made a change. But keeping track of this "changed" status was getting annoying.

I talked to @gasche and he suggested merging constant folding, constant propagation, and copy propagation into a single pass.

The plan now is to maintain an environment (of variables to constants or constant variables) and apply a `normalize` function to expressions. `normalize` would perform any substitutions it could, and then constant fold if possible. If this results in a new constant definition, then that definition would be added to the environment. `drop` instructions would then remove mappings from the environment.

In other words, constant propagation must be done "on the fly," where before, all the constant definitions would be collected upfront, and then all the substitution targets would be computed. This makes sense, since constant folding may introduce new constant definitions.

Unfortunately, this means throwing away all the previous constant folding/propagation code (which was why I avoided this idea in the first place). The code also becomes more complicated, but that seems to be unavoidable.

(I'll probably still leave the dead constant definitions alone. Deleting them messes with the instruction stream and shifts code around.)

o-:
That sounds like an excellent plan! Would the graph representation of the IR help you here? Because I am still thinking about doing this refactoring...

mhyee:
For the actual optimization, I don't think it would help. `successors_at` already gives me what I need. (Maybe once I start writing code, I'll run into a case where the graph representation helps.)

For deleting dead definitions, it would definitely help. But I'd still prefer to delegate this to a later (existing) optimization.

mhyee:
Implemented in #125.

Issue #61, opened 2017-02-21
Codemotion: sink the lifted declaration from mutable to constant variable


gasche:
See the discussions in [#58](https://github.com/reactorlabs/sourir/pull/58#issuecomment-281386063): the code-motion optimization currently creates a fresh *mutable* variable, it would be nicer to create a *constant* variable instead. This should be a relatively easy task, feel free to give it a try if you want to play with the implementation -- and to ask any question here!

The reason why this makes sense / is useful is that, in the imaginary performance model of the language, mutable declarations/assigment "cost more" than constants, as constants are intended to model short-lived stack slots (in a stack VM) of registers (in a register VM), while mutable variables invoke an indirection through the heap.

mhyee:
The basic plan:
- For all mutable declarations, wherever it is in scope, see if it gets modified.
- If not, it can be changed to a `const` declaration
- (i.e. if none of the "requires" are assignments)
- For now, we don't have functions. When we do, it becomes more difficult to reason, e.g. passing a mutable variable into a function, does it get modified?

@o- suggests looking at `hoist_assignment` in [transform](https://github.com/reactorlabs/sourir/blob/master/transform.ml), which does something very similar (in terms of looking at uses of variables). But ultimately, I think I'll put it in a separate function.

mhyee:
A motivating example: apply branch pruning (to eliminate an assignment), this allows a mutable declaration to be converted to const, then we can apply constant propagation.

mhyee:
Fixed by #91.

PR #74, opened 2017-03-27
Constant propagation

mhyee:
This PR implements constant propagation. @o- and I pair programmed the first commit, but later, I added some more tests and comments.

The strategy is pretty simple. Find all constant variable declarations of the form `const x = l` where `l` is a literal. Then, whenever `x` is in scope and used in an expression, rewrite the expression to use `l` instead of `x`. Afterwards, we can delete the declaration (and #62 does so).

The next step would be to implement constant folding (and this is why the code is in `constantfold.ml`). #73 tracks this, but I thought I'd break this up into separate PRs. Copy propagation is another thing we could implement, but I haven't really thought about it yet.

This optimization will become more powerful/useful once #61 is implemented. Then we can get some interesting examples.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,11 +30,11 @@ test_examples: sourir
 	mkdir $(TEMPDIR)/examples
 	for f in examples/*.sou; do yes 0 | ./sourir $$f --quiet \
 	  > $(TEMPDIR)/$$f.out; done
-	for f in examples/*.sou; do yes 0 | ./sourir $$f --quiet --lifetime --cm --prune \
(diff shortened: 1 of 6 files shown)


gasche:
In general I like the idea of each optimization being careful to keep the code as nice as reasonably possible; here, because we have all the information on where to remove the declaration and the drops during the transformation, I would find it natural to remove them instead of delegating to another pass. That said, if the goal is to demonstrate the interplay between several passes, the current design of course is better.

gasche:
(I reviewed the code and it is fine, feel free to merge.)

PR #75, opened 2017-04-03
Separate call instruction?

gasche:
Small change over #70; I'm not sure which is best.

### DIFF
--- a/doc/ir-semantics.text
+++ b/doc/ir-semantics.text
@@ -22,7 +22,8 @@ start     start label      (execution of segment starts here)
 a  addresses
 
 i ::=    instructions
-| const x = ce                  constant variable
+| const x = e                   constant variable
+| call x = f e*                 function call
 | mut x = e                     mutable variable
 | x <- e                        assignment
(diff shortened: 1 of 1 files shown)


doc/ir-semantics.text
```diff
@@ -106,11 +103,11 @@ Update (partial) function, returns an H:
   (P, C*) : (I, T, H, E, l) --> (P, C*) : step I(l) (P, I, T, H, E, l)
 
   (P, C*) : (I, T, H, E, l) --> (P, (C*, C')) : (I', T, H, E', start)
-    when I(l) = (const x' = call f (e, ...))
+    when I(l) = (call x = f e*)
```
o-: typo: `x'`, since it goes into continuation (last line) and should not collide with the xs from the function declaration

o- (commented):
Both is fine.

Downside of this one is that we have two introductions for constant variables.

Downside of my version is that there are two steps possible for 'const x = call'. It's not ambiguous since one is stuck in eval (it does not have `call`) but ugly nonetheless.

I slightly prefer yours

PR #76, opened 2017-04-05
refactoring: rename segment to version, remove special "active" label

o-:
This is a refactoring in preparation for the introduction of functions.
1. Fix a bug in the formalism for osr (missing function label)
2. Rename Segment to Version
3. Remove the special "active" segment. Now the active version is just
   the first version in the list. This allows us to change and add
   versions without having to rename all of them. Also the definition
   of active version is now behind the Instr.active_version api and
   should now be easier to change in the future.
4. Make the disassembler api more uniform

### DIFF
--- a/instr.ml
+++ b/instr.ml
@@ -254,8 +254,14 @@ let used_vars = function
 
 type 'a dict = (string * 'a) list
 
-type segment = instruction_stream
-type program = segment dict
+type version = label * instruction_stream
+type program = version list
+
+let active_version (prog : program) : version =
+  (List.hd prog)
+
+let replace_active_version (prog : program) (repl : version) =
+  repl :: (List.tl prog)
 
 module Value = struct
   let int n : value = Lit (Int n)
--- a/rename.ml
+++ b/rename.ml
@@ -1,5 +1,6 @@
 open Instr
 
+(* TODO unify those 3 functions *)
 let fresh_var instrs var =
   let cand i = var ^ "_" ^ (string_of_int i) in
   let is_fresh cand_var instr =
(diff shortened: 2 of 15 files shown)


instr.ml
```diff
@@ -254,8 +254,14 @@ let used_vars = function
 
 type 'a dict = (string * 'a) list
 
-type segment = instruction_stream
-type program = segment dict
+type version = label * instruction_stream
+type program = instruction_stream dict
```
gasche: Is your intent that this is equal to `(version list)`? If so, I'd rather see `version list`, or maybe define `('a, 'b) dict` and use `(label, instruction_stream) dict`.

rename.ml
```diff
@@ -21,6 +21,15 @@ let fresh_label instrs label =
...
   find 1
 
+let fresh_version_label (prog:program) label =
+  let cand i = label ^ "_" ^ (string_of_int i) in
+  let existing = fst (List.split prog) in
+  let rec find i =
+    let cand_lab = cand i in
+    if not (List.mem cand_lab existing)
+    then cand_lab else find (i+1) in
+  find 1
```
gasche, line 32: We could have a `fresh_in_list : string list -> string -> string` function to avoid repeating the concatenation logic each time.

gasche (approved):

instr.ml
o-: typo, wanted `version list`

rename.ml
o-: I'll leave it for a later, if that's ok. I add a todo...

o-:
thanks

PR #77, opened 2017-04-05
Add functions implementation

o-:
step 1 adds the function keyword and extends the parser to parse functions. toplevel instructions have to occur in one block at the beginning of the file. they get added to the implicit main function. if there are no toplevel instructions an explicit main function is expected. Similarly with versions, a function can either have a list of versions or no version keyword which creates one version called "anon". (This makes the changes backwards compatible with existing programs).

step 2 adds call / return and a bunch of tests. It also extends the scope inference to consume the formal arguments to a function as initial scope.

The implementation closely follows the formalism, except we did not model const/mut arguments there. In this version a formal argument is either annotated const or mut. In the first case we do call by value, in the second case we pass the address. This means that a function expecting a mut reference can only be called with a mut reference as argument. For example:

    call f (3) // function f (cons a)

is valid, but

    call f (3) // function f (mut a)

will cause a runtime exception. Correct version is

    mut x = 3
    call f (x) // function f (mut a)


disasm.ml
```diff
@@ -20,6 +20,16 @@ let disassemble_instrs buf ?(format_pc = no_line_number) (prog : instruction_str
     in
     format_pc buf pc;
     begin match instr with
+    | Call (var, f, exs)              ->
+      let rec dump_exs = function
+        | [] -> ()
+        | [e] -> dump_expr e
+        | e::t -> dump_expr e; pr buf ", "; dump_exs t
+      in
```
gasche: maybe for later: `let rec comma_separated dump_item = function ...` (outside this function) would factorize `dump_exs` here and `dump_vars` in the osr below.

disasm.ml
```diff
@@ -56,9 +66,17 @@ let disassemble_instrs buf ?(format_pc = no_line_number) (prog : instruction_str
   Array.iteri (dump_instr buf) prog
 
 let disassemble buf (prog : Instr.program) =
-  List.iter (fun (name, version) ->
-      Printf.bprintf buf "version %s\n" name;
-      disassemble_instrs buf version) prog
+  (* TODO: disassemble annotations *)
+  List.iter (fun {name=name; formals=formals; body=body} ->
```
gasche: `{name; formals; body}`: a single field name `f;` is syntactic sugar for `f = f` in both expressions and patterns.

eval.ml
```diff
@@ -8,7 +8,8 @@ type trace = value list
 type environment = binding Env.t
 type heap = heap_value Heap.t
 
-type status = Running | Stopped
+type status = Running | Stopped of value option
```
gasche: Maybe we could consider replacing `stop` with `return v` all the time and having `Running | Result of value`?

eval.ml
```diff
@@ -8,7 +8,8 @@ type trace = value list
...
+type continuation = variable * environment * instruction_stream * pc
```
gasche: I'm a bit surprised that an `instruction_stream` is required, but I haven't seen the code below yet.

eval.ml
```diff
@@ -128,21 +130,66 @@ let get_bool (Lit lit : value) =
...
+  let instruction =
+    if conf.pc < Array.length conf.instrs
+    then conf.instrs.(conf.pc)
+    else if conf.continuation = []
+    then Stop
+    else raise MissingReturn
+  in
+
+  match instruction with
+  | Call (x, f, exs) ->
```
gasche: `Call` and `Osr` both exceed the length of code for which easy understanding is possible while idly browsing the `reduce` function. It would be interesting to find how to split some of their logic to auxiliary functions so that the resulting code is not just a function call, it reads as a high-level description of what the function does, remaining close to the description in the ascii reduction rules.

o-:
note: I will add a typechecker tomorrow, to catch all the potential pitfalls statically:

* call to function which does not exist
* osr to function/version which does not exist
* function without return
* callsite with wrong number of args
* callsite with non-reference as call-by-name arg
* well-formedness of program (has main) and function (formal args are unique)

eval.ml
```diff
@@ -128,21 +130,66 @@ let get_bool (Lit lit : value) =
...
+    let func = Instr.lookup_fun conf.program f in
+    if (List.length exs) <> (List.length func.formals) then raise WrongNumberOfArguments else
```
gasche: I would have a mild preference for `;` instead of `else` here -- it maps the logical structure of the code better I think. (In general I decide between those options after a raise depending on the code layout, I don't have a hard rule.)

instr.ml
```diff
@@ -244,24 +262,70 @@ let used_vars = function
...
+  main : afunction;
+  functions : afunction list;
+}
+
+exception FunctionDoesNotExist of label
+exception AmbiguousFunctionName of label
+
+let lookup_fun (prog : program) (name : label) : afunction =
+  if name = "main" then prog.main else
+  match List.filter (fun {name = l} -> name = l) prog.functions with
```
gasche:
```
try List.find (fun {name = l} -> name = l)
with Not_found -> raise (FunctionDoesNotExist name)
```

(it picks the first, but then do we really need to complain at this level? We could just ignore the issue or forbid name conflicts at declaration time.)

instr.ml
```diff
@@ -244,24 +262,70 @@ let used_vars = function
...
+  | [] -> raise (FunctionDoesNotExist name)
+  | [f] -> f
+  | _ -> raise (AmbiguousFunctionName name)
+
+exception VersionDoesNotExist of label
+exception AmbiguousVersionName of label
+
+let lookup_version (func : afunction) (label : label) : version =
+  match List.filter (fun (l, _) -> label = l) func.body with
```
gasche: Again I'd use `List.find` here.

parser.mly
```diff
@@ -27,23 +27,60 @@ let scope_annotation (mode, xs) =
...
+          body = [("anon", Array.of_list (i1::instructions))];
+          annotations = [("anon", Array.of_list (a1::annotations))]; };
+      functions = fs;
+    }
+  }
+| f1=afunction fs=list(afunction)
+  {
+    let fs = f1 :: fs in
+    let main = (try List.find (fun {name = n} -> n = "main") fs with
+        | Not_found -> (Printf.printf ("missing main function\n"); exit 1)
```
gasche: `exit 1` is a bit brutal (eg. it may stop the testsuite before tests are completed, although it might protect against it). From a design point of view, I think that having this check outside the parser (in a well-formedness verification pass of its own) would be better. For now, the simpler may be to just `raise Exit` -- it will fail badly (without a nicer-for-the-user error message), but it's not a panic. Or we could declare a new exception for that, but it should not be defined in the `Parser` or `Parse` modules.

parser.mly
```diff
@@ -27,23 +27,60 @@ let scope_annotation (mode, xs) =
...
+formal_param:
+| CONST x=variable
+    { ParamConst x }
+| MUT x=variable
+    { ParamMut x }
+
+afunction:
+| FUNCTION name=variable LPAREN formals=separated_list(COMMA, formal_param) RPAREN NEWLINE optional_newlines prog=list(instruction_line)
+  {
+    let annotations, instructions = List.split prog in
```
gasche, line 68: I'm not convinced by the idea of keeping annotations and instructions separate here (it means the type-checker does not prevent you from adding one without the other). One thing you could do is to have the type of the `body` field be a type parameter, and have the parser a `(annotation * instruction_stream)`-parametrized value, and have the scope checker return just an `instruction_stream`-parametrized value.

eval.ml
o-: it is either `(function_name, version_name, pc)` or `(instructions, pc)`.
I chose the second because currently the configuration does not keep track of the current function and version name.
But maybe for the long we should switch (eg. its better for printing backtraces).

o-:
thanks for the thorough review. I agree with all points. The rest of the fixes comes tomorrow.

parser.mly
o-: I addressed this by making the version a struct and moving the annotations there. That seems a lot better.
I would really not want to go back to the situation we had before where the scope module uses a different type for the function than the rest of the system. It seems that's what you are suggesting. It's what we had before and it was a big pain at the boundaries.

gasche: That sounds good, thanks.

eval.ml
mhyee: I like calling it `Result` instead of `Stopped`, but I think it's nice to be able to return without a value. (So I'm leaning towards a rename but keeping the `option`.)

mhyee:
I looked at this but don't have much to add. Looking forward to the fixes and type checker!

eval.ml
o-: I like stop. It give some additional features. How about `stop expression`. That makes stop and return uniform.

o-:
@gasche @mhyee ptal

I decided to add the checker to the same PR as it resolves a significant portion of the issues youraised.

check.ml
```diff
@@ -0,0 +1,118 @@
...
+
+  let functions = prog.main :: prog.functions in
+
+  (* Formals args shall not contain duplicate variables *)
+  let check_formals name formals =
+    let formals = List.map (fun f -> match f with
+        | ParamConst x -> x | ParamMut x -> x) formals in
+    List.iter (fun f ->
+        if List.length (List.filter (fun f' -> f=f') formals) > 1 then
+          raise (DuplicateParameter (name, f))) formals
```
gasche: This is fine in practice, but if you want to avoid a quadratic algorithm sets are the way to go:

```ocaml
let check seen var =
  if VarSet.mem var seen
  then raise ...
  else (VarSet.add var seen)
in ignore (List.fold_left check Set.empty formals) 
```

eval.ml
```diff
@@ -128,21 +133,99 @@ let get_bool (Lit lit : value) =
...
-
 let reduce conf =
   let eval conf e = eval conf.heap conf.env e in
   let resolve instrs label = Instr.resolve instrs label in
   let pc' = conf.pc + 1 in
   assert (conf.status = Running);
-  match instruction conf with
+
+  let instruction =
+    let defaul_exit = (Simple (Lit (Int 0))) in
```
gasche: default

eval.ml
```diff
@@ -128,21 +133,99 @@ let get_bool (Lit lit : value) =
...
+    in
+    List.fold_left add Env.empty osr_def
+  in
+
+  match instruction with
+  | Call (x, f, exs) ->
+    let func = Instr.get_fun conf.program f in
+    let version = Instr.active_version func in
+    let call_env = build_call_frame func.formals exs in
+    let cont_pos = (conf.cur_fun, conf.cur_vers, pc') in
```
gasche: This is much nicer, thanks!

eval.ml
```diff
@@ -128,21 +133,99 @@ let get_bool (Lit lit : value) =
...
+        Env.add x (Const value) env
+      | Instr.ParamMut x ->
+        let get_var = function[@warning "-4"]
+          | Simple (Var x) -> x
+          | _ -> assert (false)
+        in
+        let get_addr = function
+          | Mut a as adr -> adr
+          | Const _ -> assert (false)
+        in
```
gasche: I don't like the `assert (false)` in the two functions above because they rely on the check assumption in way that feel fragile to me. That said, I thought about it and I have no obvious fix to propose (I think using a dedicated exception in this case would be nicer). I'll discuss that in a higher-level comment.

gasche (commented):
There is something with `ParamMut` that doesn't feel quite right. The code would certainly be simpler without them, but I can also see the argument for them: they let you do multi-value return with out variables.

I think that the assumption that you can just `lookup_fun` at call time may be a bit too strong. We kind of lost track of it with the current proposal, but I think we are interested in having program variables used in the call syntax, in particular to be able to model "changing the definition of base functions" in our examples -- that was at least part of the things we considered in the original language design. If we assume that in the future we won't be able to statically resolve functions at check-time, then the function call checking can be relaxed and simplified.

For `mut` parameters, an approach that I think would make sense is to use the same syntax for calls as for `osr` environments: instead of a number of parameters, function signatures would expose an "input environment", and we would use the environment-mapping syntax `const x = e, mut x = y, ...` (or even my preferred syntax `mut x <- y`) -- notice that this allows ruling out the (`ParamMut` but no `Simple (Var ..))`) case statically in the call instructoin AST structure.

Another approach is to just drop `ParamMut`, or at least the ability to alias out-variables (`ParamMut` could accept any expression as argument and only allocate it as mutable locally within the function), and add tuples for multi-value returns.

check.ml
o-: sure, but I also want to present the name of the duplicate parameter. With the sets I'd have to do it in two steps.

gasche: I think the code above does find the name of the duplicate parameter -- it's `var` available at the `raise` point.

o-:
Right, currently all the calls are statically resolved. That's also why no runtime checks for call signature and function lookup are needed. You are right that we wanted to have dynamic calls, but this needs some more thinking. In particular we need strings or symbols as values. I am happy to look into this, but I think it should happen separately. If you want I can re-add the runtime checks for function lookup and invalid argument in anticipation of non-static calls.

Then the reason I wanted to be able to pas mut by reference is because otherwise we do not have shared state. For example I believe that without this feature the effect of assignments to mut vars would be confined to the lexical scope. I assumed we wanted something more expressive with them. But I don't have a strong opinion. We can also pass all arguments by value.

Regarding using the env-rewrite syntax for calls I don't understand your idea. If we do not statically know where the call goes nothing is solved and you get additional problems (names mismatch...)

```
function bla (const x)
  ...
function main
  ...
  bla (mut y <- z)
```

If anything I would support explicit annotations for call by name / call by value at the callsite. Eg.:

```
function bla (const x)
  ...
function bla2 (mut x)
  ...
function main
  const x
  mut y
  bla (x)     # ok
  bla (&x)    # static error
  bla (y)     # ok
  bla (&y)    # dynamic error
  bla2 (x)    # dynamic error
  bla2(&x)    # static error
  bla2 (y)    # dynamic error
  bla2 (&y)   # ok
```

check.ml
o-: Ah you build it up incrementally I see, sorry

gasche:
> I wanted to be able to pass mut by reference because otherwise we do not have shared state

Ok, that's reasonable. But don't like the current syntax for the feature.

>  If we do not statically know where the call goes nothing is solved and you get additional problems (names mismatch...)

Yes: argument names become part of the function signature, and you have to check at call resolution time that the set of names is the one you expected (like we check arity).

I think there may be a benefit if it allows to share more code (and reasoning/concepts) between function calls and osr.

sourir.ml
```diff
@@ -19,104 +21,155 @@ let () =
...
+      Printf.eprintf "Program is missing an explicit or implicit main function\n";
+      exit 1
+    | Check.DuplicateFunctionDeclaration f ->
+      Printf.eprintf "Duplicate function declaration %s\n" f;
+      exit 1
+    | Check.InvalidFunctionDeclaration f ->
+      Printf.eprintf "Function %s is invalid\n" f;
+      exit 1
+    | Check.DuplicateVersion (f, v) ->
+      Printf.eprintf "Version %s in function %s is define twice\n" v f;
```
mhyee: Typo: define -> defined

o-:
> Ok, that's reasonable. But don't like the current syntax for the feature.

But do we agree that the formal parameters should say mut/const?
And you additionally want the callsite to say it too?

> ime that the set of names is the one you expected (like we check arity).

I find it overblown to be honest

sourir.ml
```diff
@@ -85,11 +85,47 @@ let () =
...
     end;
 
+    begin try Check.well_formed program with
+    | Check.MissingMain ->
+      Printf.eprintf "Program is missing an explicit or implicit main function\n";
+      exit 1
+    | Check.DuplicateFunctionDeclaration f ->
+      Printf.eprintf "Duplicate function declaration %s\n" f;
+      exit 1
+    | Check.InvalidFunctionDeclaration f ->
```
mhyee: This error (and message) isn't very helpful. Why is the function invalid?

Currently, it looks like this is only raised if the main function has a non-empty parameter list. Will this exception be used in other cases?

sourir.ml
```diff
@@ -19,104 +21,155 @@ let () =
...
+    end;
+
+    begin try Check.well_formed program with
+    | Check.MissingMain ->
+      Printf.eprintf "Program is missing an explicit or implicit main function\n";
+      exit 1
+    | Check.DuplicateFunctionDeclaration f ->
+      Printf.eprintf "Duplicate function declaration %s\n" f;
+      exit 1
+    | Check.InvalidFunctionDeclaration f ->
```
mhyee: This error (and message) isn't very helpful. Why is the function invalid?

Currently, it looks like this is only raised if the main function has a non-empty parameter list. Will this exception be used in other cases?

check.ml
```diff
@@ -0,0 +1,118 @@
+open Instr
+
+exception MissingMain
+exception DuplicateFunctionDeclaration of identifier
+exception InvalidFunctionDeclaration of identifier
+exception DuplicateVersion of identifier * label
+exception EmptyFunction of identifier
+exception DuplicateParameter of identifier * variable
+
+exception ErrorAt of identifier * label * exn
```
mhyee: Minor detail, but I didn't find it obvious why this was called `ErrorAt` and why only certain exceptions are wrapped by `ErrorAt`. Only by looking at usages of `ErrorAt` did I realize it's for errors at a specific function and version (as opposed to general errors in the program).

I think a short comment here would help.

gasche:
> But do we agree that the formal parameters should say mut/const?

If we keep the aliasing feature, sure.

> I find it overblown to be honest

Our tastes disagree; I clearly find `x <- y` more elegant than `&y`. But if we don't agree on the former, I'm willing to accept the later, it's a good improvement over your current proposal. `&y` would only be valid for `mut` variable (checked statically as you mention) and only be used for aliasing in function and osr calls -- and `mut x` binders would only accept arguments of this shape.

check.ml
```diff
@@ -0,0 +1,118 @@
...
+    List.iter (fun f ->
+        if List.length (List.filter (fun f' -> f=f') formals) > 1 then
+          raise (DuplicateParameter (name, f))) formals
+  in
+
+  let check_version func version =
+    let instrs = version.instrs in
+
+    if func.name <> "main" then
+    begin if Array.length instrs = 0 then raise MissingReturn else
```
mhyee: I liked @gasche's [earlier comment](https://github.com/reactorlabs/sourir/pull/77#discussion_r109983318) about using `;` instead of `else`. It's a lot like writing asserts/ifs-without-else at the beginning of C functions.

mhyee:
General notes:

1) I'd like to see a few more examples added, especially with const and mut parameters. But this will have to wait until we agree on the syntax.

2) The implementation has diverged from the semantics document. The semantics don't include mutable parameters. Also, the structure of continuations has changed.

---

I prefer @o-'s updated proposal for syntax. While `&y` may be uglier and less clear than `x <- y`, I don't like making argument names part of the function signature. It also raises a question about whether the order of arguments matters.

o-:
@mhyee to your second comment: yes we ignored the issue in the formalism. I'll update.
@gasche ok, I'll go for & annotation. I find the renaming to complicated

o-:
ptal
(everything should be addressed except dynamic lookup, which I'll start in a separate PR tomorrow)

check.ml
```diff
@@ -0,0 +1,123 @@
+open Instr
+
+(* Those are exception at the scope of the program *)
```
mhyee: Nitpicking typos: Th**e**se are exception**s**

check.ml
```diff
@@ -0,0 +1,123 @@
...
+exception MissingMain
+exception InvalidMain
+exception DuplicateFunctionDeclaration of identifier
+exception DuplicateVersion of identifier * label
+exception EmptyFunction of identifier
+exception DuplicateParameter of identifier * variable
+
+(* The following are exception at the scope of a particular
```
mhyee: Nitpicking typos: The following are exception**s** ... they will be wrapped in a**n** ErrorAt exception.

examples/fun_mut_const.sou
```diff
@@ -0,0 +1,16 @@
...
+call c3 = take_mut (&x)
+
+const s1 = (c1+c2)
+print (s1+c3)
+
+
+function take_const (const a)
+  return a
+
+function take_mut (mut a)
```
mhyee, line 15: Show an example of assigning to a mutable variable? (Then again, I think we already have this in a test.)

constantfold.ml
```diff
@@ -28,6 +27,17 @@ let const_prop prog =
   let convert x l instr =
     let replace = Replace.var_in_exp x l in
     match instr with
+    | Call (y, f, es) ->
+      assert (x <> y);
+      let replace_arg = function
```
mhyee: Should we move this to `replace.ml` (since `rename.ml` has a similarly short function for renaming in an arg)?

mhyee:
Looks fine. I agree that further work (beyond addressing review feedback) should go in a new PR. This one is starting to get large.

o-:
@gasche 
> But if we don't agree on the former, I'm willing to accept the later, it's a good improvement over your current proposal. 

The reason I don't like the named arguments proposal is that while I think it's a nice feature for a high-level language I really don't see the point in a language which essentially is a compile target.

But just to be clear I absolutely agree to the second part of your sentence. It is much better to statically indicate which semantics you want at the call site.

gasche:
I merged this to make #79 easier to review. Thanks for the impressive amount of work overall, including in the last "address the review" commit.

It wouldn't feel real if I did not have some complaint: I don't like (val/var) to replace (const/mut), because that I think that "variable" should not be captured to mean "mutable" (constant variables are perfectly fine and in fact they are better if you ask me in general). I liked (const/mut), but if you want something else, maybe (val/ref)?

o-:
> It wouldn't feel real if I did not have some complaint: I don't like (val/var) to replace (const/mut)

hehe. I touched it because it was inconsistent. Which one i chose was rather mindless. I like (val/ref). I'll look into it.

PR #78, opened 2017-04-07
fix typo

o-:
or was this an intentional thing? I would even be ok with leaving
it in there. I find it neat.


gasche:
It's 100% a typo -- I guess it's neat because it comes from French :-'

o-:
yeah and it was so consistently littered all over the place :)

PR #79, opened 2017-04-07
function references as values and indirect calls

o-:
Based on #77 

Adds to the language:
```
call x = *e ()  expects f to evaluate to a function reference
call x = f ()   is a static call to the function called f
const f = &&n   assigns a function reference to the function called n to the variable f
```

Internally they are called Call and StaticCall

Notes:
* This PR departs from Value = Literal in the implementation. I find it a bit cleaner this way but I don't have a strong opinion in case you liked it the way it was before.
* The syntax to create a function reference (`&&f`) is not super nice. Better suggestions are welcome. `&f` does not work due to ambiguity with the static call-by-ref/call-by-value annotation.
* Maybe we want `call x = e ()` and `static call x = f ()` at the surface instead of `call x = *e ()` and `call x = f ()`?

### DIFF
--- a/analysis.ml
+++ b/analysis.ml
@@ -12,7 +12,7 @@ let successors_at (instrs : instruction_stream) pc : pc list =
   let all_succ =
     match instr with
     | Decl_const _ | Decl_mut _ | Assign _ | Drop _ | Clear _ | Read _
-    | Call _ | StaticCall _ | Label _ | Comment _ | Osr _ | Print _ ->
+    | Call _ | Label _ | Comment _ | Osr _ | Print _ ->
       let is_last = pc' = Array.length instrs in
       if is_last then [] else [pc']
     (* those are the instructions which manipulate controlflow:  *)
--- a/analysis.ml
+++ b/analysis.ml
@@ -12,7 +12,7 @@ let successors_at (instrs : instruction_stream) pc : pc list =
   let all_succ =
     match instr with
     | Decl_const _ | Decl_mut _ | Assign _ | Drop _ | Clear _ | Read _
-    | Call _ | Label _ | Comment _ | Osr _ | Print _ ->
+    | Call _ | StaticCall _ | Label _ | Comment _ | Osr _ | Print _ ->
       let is_last = pc' = Array.length instrs in
       if is_last then [] else [pc']
     (* those are the instructions which manipulate controlflow:  *)
--- a/doc/ir-semantics.text
+++ b/doc/ir-semantics.text
@@ -5,7 +5,7 @@ F ::= (lᵥ ↦ V)*                          function body:      can contain mul
 V ::= start ↦ i, I                       a function version: instruction stream with dedicated start
 I ::= (l ↦ i)*                           instruction stream: labeled instructions
 
-Df ::= Function f(mod x, ...)          function declaration
+Df ::= Function f((mod x)*)          function declaration
 
 x        variables
 
@@ -25,7 +25,7 @@ a  addresses
 
 i ::=    instructions
 | const x = e                   constant variable
-| call x = f arg*               function call
+| call x = fun (arg*)           function call
 | mut x = e                     mutable variable
 | x <- e                        assignment
 | branch e l₁ l₂                conditional
@@ -38,9 +38,14 @@ arg ::=
 | e
 | &x
 
+fun ::=
+| *e
+| f
+
 e ::=    (simple) expressions
 | lit                   literals
 | x                     variables
+| &&f                   function reference
 | primop(x, ...)        primitive operation (pure)
 
 lit ::=  literals
@@ -93,21 +98,24 @@ Update (partial) function, returns an H:
 
 ## Evaluation of simple expressions:
 
-  eval H E x = (H,E)[x]
-  eval H E lit = lit
-  eval H E primop(v, ...) = 〚primop〛(v, ...)
+  eval P H E x = (H,E)[x]
+  eval P H E lit = lit
+  eval P H E primop(v, ...) = 〚primop〛(v, ...)
+  eval P H E &&f            = f
+     and Df ∈ dom P
+     and Df  = Function f (mod y)*
 
 ## Osr transformation of environment
 
-  osr-env H E (const x = e) = (x ↦ eval H E e)
-  osr-env H E (mut x)       = (x ↦ ⊥)
-  osr-env H E (mut x = y)   = (x ↦ a)
+  osr-env P H E (const x = e) = (x ↦ eval P H E e)
+  osr-env P H E (mut x)       = (x ↦ ⊥)
+  osr-env P H E (mut x = y)   = (x ↦ a)
     where a = E(y)
 
 ## Argument evaluation
 
-  eval-arg H E const  e =  val eval H E e
-  eval-arg H E mut   &x =  ref a
+  eval-arg P H E const  e =  val eval P H E e
+  eval-arg P H E mut   &x =  ref a
     where E ∋ (x ↦ ref a)
 
 ## Reduction relation '-->'
@@ -117,14 +125,23 @@ Update (partial) function, returns an H:
   (P, C*) : (I, T, H, E, l) --> (P, (C*, C')) : (I', T, H, E', start)
     when I(l) = (call x = f (arg*))
      and Df ∈ dom P
-     and Df  = Function f (mod y)*
+     and Df  = Function f ((mod y)*)
      and I' := hd P(Df)
-     and E' := (y ↦ eval-arg H E mod arg)*
+     and E' := (y ↦ eval-arg P H E mod arg)*
+     and C' := (x, E, I, succ I l)
+
+  (P, C*) : (I, T, H, E, l) --> (P, (C*, C')) : (I', T, H, E', start)
+    when I(l) = (call x = *ef (arg*))
+     and f  := eval P H E ef
+     and Df ∈ dom P
+     and Df  = Function f ((mod y)*)
+     and I' := hd P(Df)
+     and E' := (y ↦ eval-arg P H E mod arg)*
      and C' := (x, E, I, succ I l)
 
   (P, (C*, C')) : (I, T, H, E, l) --> (P, C*) : (I', T, H, E'', l')
     when I(l) = (return e)
-     and v   := eval H E e
+     and v   := eval P H E e
      and C'   = (x', E', I', l')
      and E'' := E'[x' ↦ v]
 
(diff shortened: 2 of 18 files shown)


gasche:
We already have a definition `type value = Literal of literal`,  why doesn't it work to add a new constructor there?

(I merged the previous PR but it seems they don't completely agree on history. Can you rebase?)

o-:
> We already have a definition type value = Literal of literal, why doesn't it work to add a new constructor there?

Yes, but now I wanted `FunRef of string` as the literal and `FunRef of afunction ref` as the value. If I kept the `Literal` in the `Value` type then we'd have both in there. For me it was a bit of a conceptual mismatch in the first place how literals where wrapped in values. I think there should be a mapping between values and literals, but I don't think that a literal is a subset of values.

Let me quickly rebase...

eval.ml
```diff
@@ -84,59 +85,63 @@ let clear heap env x =
...
+    | Plus, [v1; v2] -> Int (value_plus v1 v2)
     | (Eq | Neq | Plus), _vs -> raise (Arity_error op)
     end
 
-let get_int (Lit lit : value) =
-  match lit with
-  | Int n -> n
-  | (Nil | Bool _) as other ->
-     let expected, received = Int, literal_type other in
+let get_bool (v : value) =
```
mhyee, line 123: Why was `get_int` removed?

eval.ml
```diff
@@ -27,11 +27,12 @@ type configuration = {
   continuation : continuation list;
 }
 
-type literal_type = Nil | Bool | Int
-let literal_type : literal -> literal_type = function
+type value_type = Nil | Bool | Int | FunRef
```
mhyee: `FunRef` -> `Fun_ref`
(Might as well start nitpicking the constructor names.)

doc/ir-semantics.text
```diff
@@ -93,21 +98,24 @@ Update (partial) function, returns an H:
 
 ## Evaluation of simple expressions:
 
-  eval H E x = (H,E)[x]
-  eval H E lit = lit
-  eval H E primop(v, ...) = 〚primop〛(v, ...)
+  eval P H E x = (H,E)[x]
+  eval P H E lit = lit
+  eval P H E primop(v, ...) = 〚primop〛(v, ...)
+  eval P H E &&f            = &&f
```
mhyee: Should the second `&&f` be just `f`? (Maybe choosing a different letter, e.g. `g`, would be clearer.)

doc/ir-semantics.text
```diff
@@ -93,21 +98,24 @@ Update (partial) function, returns an H:
...
+     and Df ∈ dom P
+     and Df  = Function f (mod y)*
```
mhyee: The definition for `Df` uses the syntax `Function f(mod x, ...)`. I guess it would be `Function f(mod y, ...)` in this case.

doc/ir-semantics.text
```diff
@@ -25,7 +25,7 @@ a  addresses
 
 i ::=    instructions
 | const x = e                   constant variable
-| call x = f arg*               function call
+| call x = fun arg*             function call
```
mhyee: `fun (arg*)`.

analysis.ml
```diff
@@ -12,7 +12,7 @@ let successors_at (instrs : instruction_stream) pc : pc list =
   let all_succ =
     match instr with
     | Decl_const _ | Decl_mut _ | Assign _ | Drop _ | Clear _ | Read _
-    | Call _ | Label _ | Comment _ | Osr _ | Print _ ->
+    | Call _ | StaticCall _ | Label _ | Comment _ | Osr _ | Print _ ->
```
mhyee: `Static_call`

mhyee:
> Internally they are called Call and StaticCall

Maybe we should have `Static_call` and `Dynamic_call`? That makes it more explicit, since `Call` by itself doesn't give you much information.

> This PR departs from Value = Literal in the implementation. I find it a bit cleaner this way but I don't have a strong opinion in case you liked it the way it was before.

I like this change.

> The syntax to create a function reference (`&&f`) is not super nice. Better suggestions are welcome. `&f` does not work due to ambiguity with the static call-by-ref/call-by-value annotation.

Yeah, I really don't like that. `&x` isn't the clearest syntax, but at least it resembles C/C++. Unfortunately, `&&f` is also from C++ but means something completely different.

We could choose some other symbol, but if we're fine with adding new keywords, how about `const f = funref foo`? The advantage is that this is clearer. But then we could get something like `call x = *(funref foo) ()`, which is slightly better than `call x = *&&foo ()`.

> Maybe we want `call x = e ()` and `static call x = f ()` at the surface instead of `call x = *e ()` and `call x = f ()`

Again, I would prefer the syntax to be more explicit, so `dynamic call x = e ()` and `static call x = f ()`.

gasche:
I still fine the `literal` change suspicious. After looking a bit more at your change, I understand part of your point: `literal_type` is wrong (it is not semantically related to literals but to any function on which primitives act), I should have called it `type_tag` or `dynamic_type` from the start (`value_type` is not bad but it doesn't indicate that this is only partial type information, not full static information). Please rename it, and feel free to add a `Funref` case to it, it makes perfect sense. (The `Funref` tag could carry the function arity, which would turn arity-mismatch into a dynamic type error, or not, as you prefer.)

On the other hand, "literal" *is* a reasonable class to have for both terms and values of a language, it describes the class of ground terms/values that can be written down as atomic... literals. This patch completely duplicates the definition of `literal` within `value`, and I don't see the point.

The question of whether a function *name* should be a literal (as in your proposed implementation) or not is interesting. It is usually not the case in languages that I know of, but it is correct to consider function names as literal here as they are always defined by a statically-known set of possible functions. So I think that two different designs would be acceptable:

```ocaml
and simple_expression =
  | Lit of literal
  | Fun_ref of fun_ref
  | Var of variable
and literal =
  | Nil
  | Bool of bool
  | Int of int
```

or the one you propose:

```ocaml
and simple_expression =
  | Lit of literal
  | Var of variable
and literal =
  | Nil
  | Bool of bool
  | Int of int
  | Fun_ref of fun_ref
```

(The second is less standard but justified in absence of local function declarations.)

Values, unlike literal, are not defined by a role in the term syntax, but by the semantic of being the possible results of evaluating an expression. The change you propose (to duplicate the definition of literals within the definition of values) is not semantically wrong in this sense, but it introduces duplication (also in term definitions, see [check.ml:93-94](https://github.com/o-/sourir/blob/f288dafc06b5cd0669f799b6c620bfa5583d9b98/check.ml#L93-L94) for example.

Finally, I find the type of function *values* that you introduced fairly fishy. I don't see why it would be a mutable reference (`ref`). Why not just keep function names as the evaluated values, if all function declarations are global? (If you keep function names as literal as you do, the change comes from free from keeping the `Lit of literal` case, otherwise you can add it to the type of values just as we added a function case to the type of expressions.)

gasche:
Let me do the Jan here: do we really need the distinction between static calls and non-static calls? We could have distinction between variable names and function names as a lexical conventions (for example: function names must start with an uppercase letter), and just have dynamic calls without any particular symbol. Then it is of course fine to optimize `call x = Foo(...)` into a different semantics than `call x = foo(...)` to remove the lookup, but that could also be considered premature optimization. (Also not distinguishing the two instruction means that constant-folding `const foo = Foo; call x = foo(...)` will immediately enable whatever sort of inlining we do for statically-known calls, without requiring inlining to lower the instruction or a separate pass just for that.)

o-:
Update: ok, I finally understand the literal/value confusion thanks to your explanation. I have added a new commit which removes the 'Literal' type since we do not seem to agree on it's definition. Instead I call it `tagged_value` at runtime and `Constant of tagged_value` in the expression.

> On the other hand, "literal" is a reasonable class to have for both terms and values of a language

That's basically what I am skeptical about. Actually it feels quite wrong. From my point of view there are some runtime values which have a literal *representation*. But the literal itself won't be there at runtime. The value `int(1)` (ie. the type tag 'int' with the bitpattern for 1) has the literal representation `"1"` in the source code -- or maybe `literal(int, 1)` in the parse tree. But a value (as in the thing that exists at runtime) with a type of literal (as in the textual represenation of that value) does not make sense. The only possible way for it to exist is for example in lisp `quote(1)` which I would describe as creating the *value* `literal(int(1))`.

> The question of whether a function name should be a literal (as in your proposed implementation) or not is interesting.

maybe I am just confused again, but I (or my implementation) never claims that a function name is a literal. The expression `&&bla` is the literal representation of the function *reference* to bla.

> Why not just keep function names as the evaluated values, if all function declarations are global?

I wanted to say that `&&foo` creates a function reference and `call *x` invokes the function stored (by ref) in `x`. The other way round it's more like `&&foo` creates a symbol and `call *x` looks up the symbol stored in x and calls the function with the name of that symbol (if it exists). Those two are only equivalent if every function has a globally unique identifier and you can only create symbols of valid function names. But as you say those two conditions are fulfilled in our case. For the sake of simplicity of the implementation I changed it to `Fun_ref of string`. Because the correct solution to my approach would involve building up the table of functions beforehand.

> Finally, I find the type of function values that you introduced fairly fishy. I don't see why it would be a mutable reference

Agreed, mutable certainly doesn't make sense.

> Let me do the Jan here: do we really need the distinction between static calls and non-static calls? 

No, not at all. Initially I had only call. Update: How about ditching static call. Then all `call` instructions are indirect (alas the `*` is not needed anymore). A static call is now `call x = &&func ()` and a dynamic one is `call x = func ()`.

> Yeah, I really don't like that. &x isn't the clearest syntax, but at least it resembles C/C++. Unfortunately, &&f is also from C++ but means something completely different.

mhm, problem is I did not find a better symbol :) `&&` is also used to refer to the address of goto labels in gcc, that was my association.

eval.ml
o-: dead code

doc/ir-semantics.text
o-: right. Can't take g because `&&f` and `function *f* ()` need to correspond.

gasche:
(You haven't given an opinion on the idea to have a lexical distinction between variable and function names, to avoid having `Fun_ref` explicit in the source syntax.)

o-:
> (You haven't given an opinion on the idea to have a lexical distinction between variable and function names, to avoid having Fun_ref explicit in the source syntax.)

right, I think what would be nice is that we can remove the `&&`. I would still keep the `Constant (Fun_ref f)` because conceptually `&&f` is supposed to statically refer to the function f whereas `f` is a dynamic lookup.

see:
```
mut f = &&f
call f
call &&f
f <- &&g
call &&f
call f
```

What mostly worries me is that at some point we will have more than tw sorts of variables and the convention doesn't scale. Also it's a bit confusing to have `F` early bound but `f` late bound.

gasche:
Yes, the internal representation would remain the same, it's just that `F` in expression context would elaborate to `Fun_ref "F"` instead of `Var "F"`.

o-:
As I said before I would prefer a better symbol. Maybe `'`?
```
call x = f ()
call x = 'f ()
```
doesn't look too bad.

On the other hand I don't have a strong opinion and would settle for either (unquote or upper case).

gasche:
Fine with `'f`.

gasche:
> What mostly worries me is that at some point we will have more than tw sorts of variables and the convention doesn't scale.

Let's think about this later. With the power to break backward-compatibility we won't be blocked.

> Also it's a bit confusing to have F early bound but f late bound.

You don't have to consider F early bound, it is just a variable in a different namespace. You find it early because of the global guarantee.

Re. `literal`, `value` etc.: I am still not convinced, but I think it is time to move forward in the interest of making progress, so I would suggest that we go along with your current proposal for now. (`tagged_value` is rather heavy if there are no other kinds of values, maybe just `value` is reasonable?)

If you make whatever aesthetic change you like, drop `&&` and merge, I can implement the lexer part. Of course feel free to do it yourself -- it's easy.

o-:
Ok, so I changed tagged_value to value as a fixup and swapped `'` for `&&`. The lexer and parser are updated too.

Though I must say I like the namespace explanation and am still open to ditch `'` in favor of this version in the future.

As for the literal discussion I really think the problem is that we do not agree on the definition of what a literal is. Imho a literal only exists in source code (maximally in a parse tree). But since there are adequate alternative terms we do not disagree on (string representation and constant value), I suggest we use those.

gasche:
Yes; the fact that the current proposal imposes that the "value" representation of functions is the same that the "source" representation of functions be the same is a bit unfortunate, but on the other hand it works for the time being.

gasche:
Merged, thanks for the hard work!

PR #80, opened 2017-04-07
drive-by fix some names

o-:
* bindings are either (x -> val v) or (x -> ref a)
* formal parameters either take a 'mutable reference' or a 'const value'
* actual arguments are passed by_value or by_name
* the (slight majority) convention for constructor names in our code is Camel_dash

### DIFF
--- a/check.ml
+++ b/check.ml
@@ -45,7 +45,7 @@ let well_formed prog =
   (* Formals args shall not contain duplicate variables *)
   let check_formals name formals =
     let formals = List.map (fun f -> match f with
-        | ParamConst x -> x | ParamMut x -> x) formals in
+        | Const_val_param x -> x | Mut_ref_param x -> x) formals in
     let check seen var =
       if VarSet.mem var seen
       then raise (DuplicateParameter (name, var))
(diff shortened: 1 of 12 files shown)


gasche:
I've been intentionally not commenting on your horrible use of CamelCase in names (in OCaml the underscore is king, although some people import evil habits such as yours from other languages), but I welcome the change.

Another thing that is related: the language supports having several types with constructors of the same name (you use typing annotations to disambiguate when necessary). This means that we can avoid having redundancy in constructor name (recently you introduced `AtLeastScope` and `ExactScope` for example, I think the `Scope` part is unnecessary). Constructor name clashes *do* come with a bit of ergonomic overhead (you get warnings when there is not enough type information), so it's better when the constructor names naturally end up being different, but I would still avoid prefixing in all cases.

gasche:
Should we wait for your function work to calm down to merge this to minimize conflicts, or go ahead now?

o-:
> Another thing that is related: OCaml support having several types with constructors of the same name 

In the case of scope I added it because I moved it to the "wrong" module `instr.ml` to avoid circular dependency. I know that the correct solution would have been to create .mli files. But luckily it evaded being spotted by your trained reviewer eye :)

You can point out the CamelCase in the future so we eventually get consistent again. (though I must say This_looks_very weird to me)

> Should we wait for your function work to calm down to merge this to minimize conflicts, or go ahead now?

I can rebase...

PR #81, opened 2017-04-10
fix semantics: remove static call

o-:
In #79 it was decided to remove static calls and instead use the
verbose version `call x = 'f ()` where `'f` statically refers to
the function named `f`.

+ some drive-by fixes


o-:
To be reviewed...

(since it's just typos and negligence I merge it before review. Tell me if you find it not appropriate)

gasche:
All good.

> Tell me if you find it not appropriate.

I find it perfectly appropriate.

PR #83, opened 2017-04-10
[WIP] add named tuples

o-:
Let's talk about how we add tuples. This PR is intended to start a discussion about the design. I volunteer to implement. But if anybody else wants to continue from here it can also be limited to just the changes to ir-semantics.text.

The current proposal is very lightweight. With this we could type-check const tuples but not mut. Also update gives no guarantees preserving types. This is kind of in line with the rest of the language since we don't have type annotations except for mut/const. (But it is debatable...)

Also I tried to keep eval non-recursive (though I am not sure anymore why we wanted that). You can create recursive tuples step-by-step:

```
const x3 = <a:3, b:nil>
const x2 = <a:2, b:x3>
const x1 = <a:1, b:x1>
```

Access works as you expect.

I also added strings. Here is an object system on top of it:

```
function lookup (mut obj, const selector)
    mut m = obj.vtable
  loop:
    branch (m.selector == selector)  done next
  next:
    m <- m.next
    goto loop
  done:
    return m.func

function invoke (mut obj, const selector, const args)
  call m = 'lookup (obj, selector)
  call r = m (obj, args)
  return r

function vec_length (mut this)
  return sq(this.x*this.x + this.y*this.y)

function vec_add (mut this, const args)
  const other = args.1
  call res = 'new_vec (this.x+other.x, this.y+other.y)
  return res

function new_vec (const x, const y)
  const vtable =
   <selector:"add", method:'vec_add, next=
    <selector:"lenght", method='vec_length, next=<>>>
  return <vtable:vtable, x:x, y:y>

function main ()
  call a = 'new_vec (1,2)
  call b = 'new_vec (2,1)
  call ab = 'invoke (&a, "add", <1:b>)
  call l = 'invoke (&ab, "length")
```

Also, we could remove nil from the language since we have the empty tuple `<>` now.

### DIFF
--- a/doc/ir-semantics.text
+++ b/doc/ir-semantics.text
@@ -28,11 +28,14 @@ i ::=    instructions
 | call x = e (arg*)             function call
 | mut x = e                     mutable variable
 | x <- e                        assignment
+| x.n <- e                      tuple update
 | branch e l₁ l₂                conditional
 | goto l                        goto
 | print e                       print
 | return e                      return
 | osr(e, f, lᵥ, l, osr-map*)    osr : (can be used for osr-in and osr-out)
+
+n - tuple index name
 
 arg ::=
 | e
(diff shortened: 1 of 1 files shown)


o-:
let's shelve this for now

mhyee:
We decided to go with arrays (#85) for now.

PR #84, opened 2017-04-10
update semantics to match evaluation by the the implementation

o-:


doc/ir-semantics.text
```diff
@@ -38,11 +38,14 @@ arg ::=
...
+e ::=     expression
+| se                    simple expression
+| primop(se*)           primitive operation (pure)
+
+se ::=    simple expressions
 | lit                   literals
 | x                     variables
-| 'f                    function reference
-| primop(x, ...)        primitive operation (pure)
+| 'f                    create function reference
```
gasche, line 48: would maybe `literal` be a short name for "simple expressions"? ;-)

o-: no because in the implementation they are called simple expressions :)

PR #85, opened 2017-04-10
add arrays

o-:
Maybe this is what we want instead of #83

### DIFF
--- a/doc/ir-semantics.text
+++ b/doc/ir-semantics.text
@@ -28,6 +28,7 @@ i ::=    instructions
 | call x = e (arg*)             function call
 | mut x = e                     mutable variable
 | x <- e                        assignment
+| x[e] <- e                     array assignment
 | branch e l₁ l₂                conditional
 | goto l                        goto
 | print e                       print
@@ -41,6 +42,10 @@ arg ::=
 e ::=     expression
 | se                    simple expression
 | primop(se*)           primitive operation (pure)
+| array(se)             empty array constructor
+| [se*]                 array constructor
+| x[se]                 array index
+| length(x)             array length
 
 se ::=    simple expressions
 | lit                   literals
@@ -55,6 +60,7 @@ lit ::=  literals
 v :=     values
 | lit                   literals
 | f                     function reference
+| [v*]                  array
 
 (Note: heap adresses are not values)
 
@@ -107,7 +113,18 @@ Update (partial) function, returns an H:
 
   eval P H E se          = simple-eval P H E se
   eval P H E primop(se*) = 〚primop〛(v*)
-    where v := simple-eval P H E se
+    where (v := simple-eval P H E se)*
+  eval P H E [se*]       = [v*]
+    where (v := simple-eval P H E se)*
+  eval P H E x[se]       = vₙ
+    where v := (H,E)[x]
+      and n := simple-eval P H E se
+      and v  = [v₁..., vₙ, ...]
+  eval P H E array(se)   = [nil⁽ⁿ⁾]
+    where n := simple-eval P H E se
+  eval P H E length(x)   = n
+    where v := (H,E)[x]
+      and v  = [v₁..., vₙ]
 
 ## Environment actions (variable binding, function call, osr)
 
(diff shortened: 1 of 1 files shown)


doc/ir-semantics.text
```diff
@@ -92,15 +102,29 @@ Update (partial) function, returns an H:
...
-  eval P H E 'f             = f
+  simple-eval P H E x   = (H,E)[x]
+  simple-eval P H E lit = lit
+  simple-eval P H E 'f  = f
     where Df ∈ dom P
       and Df  = Function f (formal*)
 
+  eval P H E se          = simple-eval P H E se
+  eval P H E primop(se*) = 〚primop〛(v*)
+    where v := simple-eval P H E se
```
gasche: I'd write `(v := simple-eval P H E se)*` for consistency -- same below.

doc/ir-semantics.text
```diff
@@ -38,11 +39,18 @@ arg ::=
...
 | &x
 
-e ::=    (simple) expressions
+e ::=     expression
+| se                    simple expression
+| primop(se*)           primitive operation (pure)
+| array(se)             empty array constructor
+| [se*]                 array constructor
+| x[se]                 array index
+| |x|                   array length
```
gasche: I'd just use `length(array)`, but it's superficial. Otherwise the proposed primitives look good.

doc/ir-semantics.text
```diff
@@ -176,6 +193,14 @@ extended with the new heap fragment into a new heap `H'`.
        = (I, T, (H,E)[x ← v], E, succ I l)
     when v := eval P H E e
 
+  step   (x[eᵢ] ← eₐ) (P, I, T, H, E, l)
+       = (I, T, (H,E)[x ← v'], E, succ I l)
+    when vᵢ := eval P H E eᵢ
```
mhyee, line 198: Shouldn't this be `n` and not `v_i`? (See line 201.)

o-: good catch

o-: Directly pushed the fix

gasche:
After yesterday's meeting it now appear to me that it is more natural to allocate arrays on the heap. I would define the array values as `array(a)`, where `a` is a heap location pointing to the array `[v*]`. This ensures that we have the correct aliasing semantics when passing an array to a function. Unless someone objects, I will probably do the change with @aviralg.

gasche:
( @o- do you agree in principle? )

o-:
@gasche is it a problem that we can have arrays as values? My understanding of the current semantics was that if an array is declared `const` the `a[i] <- x` get's stuck (since it only covers arrays on the heap). And passing a mut array by value of course has to create a copy. So I am not sure what you mean by
> This ensures that we have the correct aliasing semantics when passing an array to a function.

gasche:
I had a by-reference semantics in mind (rather, I would consider that the value of an array is a pointer to a heap-allocated array, making the by-value and by-reference semantics coincide). I'm skeptical about a semantics that imposes copies of arrays. On the other hand, what we have now works and is consistent (I hadn't realized when giving a read to the proposal that writes to `const` arrays would get stuck), so it would be a different design rather than fix.

gasche:
(To elaborate: I consider that a `const x = array(n)` would create a variable that will always point to the same array, but the array elements could change. `mut x = ...` lets you change which array a variable refers to.)

(We don't have a collection/refcounting mechanism for arrays, but then we have no collection for the heap in general so it's not a new concern, we just decided to not model that.)

o-:
On 04/25/2017 05:21 PM, Gabriel Scherer wrote:
> I had a by-reference semantics in mind (rather, I would consider that the value of an array is a pointer to a heap-allocated array, making the by-value and by-reference semantics coincide).

ok, does that mean that for function calls only &a would be a valid
argument?

Note that this goes against the current strategy of constant folding
muts by doing a mut-to-const pass.

> I'm skeptical about a semantics that imposes copies of arrays.

It's a bit ugly but not unheard of. (C++ copy ctrs).

We could also say that you can have const arrays, but you cannot pass
them as arguments. That could potentially give us the best of both worlds.

> On the other hand, what we have now works and is consistent (I hadn't realized when giving a read to the proposal that writes to `const` arrays would get stuck)

Yeah that could maybe be made a bit more explicit. (same goes for the
normal assign case)

> so it would be a different design rather than fix.

right. I am not sure which is better. Maybe just try the alternative
approach and we'll see what feels better?

gasche:
> ok, does that mean that for function calls only &a would be a valid argument? 

No, if the value of an array is always of the form `array(a)`, both `const` and `mut` have consistent behavior (in particular `array(a)` is sometimes found on the heap itself, when it is the current value of a mutable binding) and all modes of argument-passing work.

>  Note that this goes against the current strategy of constant folding muts by doing a mut-to-const pass.

In other words, you cannot reason on the element values of constant array when doing a constant folding (even of const) pass. But you can rely on their length.

PR #86, opened 2017-04-10
change osr syntax and instruction

o-:
This fell out of a different (abandoned) change. Since imho the syntax slightly improved I am proposing the following change.

Syntax is new `osr e (f,v,l) [map]` instead of `osr e f v l [map]`. Also the internal representation uses a struct instead of a tuple to make matching a bit more intuitive.


o-:
I might still change this one. no need to look at it.

PR #87, opened 2017-04-11
Cleanup transformations

o-:
I am sorry, this PR got out of hands. It contains the following 3 changes:

1. unify the api for transformations
2. change the types to give analysis and transformation access to the formal parameters of a function
3. fix broken optimizations which did not care about aliases introduced by function call


o-:
btw. the first commit mostly moves stuff around. The only additions are to transform.ml

sourir.ml
```diff
@@ -16,10 +16,14 @@ let () =
...
-    let constprop = Array.mem "--prop" Sys.argv in
-    let lifetime = Array.mem "--lifetime" Sys.argv in
+    let rec find_opts = function
+      | [] -> []
+      | o::r when String.length o > 5 && String.sub o 0 6 = "--opt=" ->
+        let opts = String.sub o 6 ((String.length o) - 6) in
+        String.split_on_char ',' opts
+      | o::r -> find_opts r
+    in
+    let optimizations = find_opts (Array.to_list Sys.argv) in
```
gasche: I think it would be worth it to have a look at the documentation of the `Arg` module for how to actually do this properly. (I never had the time to squash that particular insect.)

(There are even fanciers/nicer way than Arg, in particular the [Cmdliner](http://erratique.ch/software/cmdliner) library which is extremely nice, but has a higher entry cost and I think that's not where we want to spend our sophistication right ow.)

transform.ml
```diff
@@ -1,205 +1,103 @@
...
-    annotations = None } in
-  { func with
-    body = cleanup :: func.body }
+let (!!) t instrs =
+  match t instrs with
+  | None -> false, instrs
+  | Some instrs -> true, instrs
 
+let (!!!) t instrs =
+  let _, instrs = !! t instrs in instrs
```
gasche: I'd rather avoid prefix operators used in a non-local way (shortening stuff within a single function is fine, but otherwise...). What about `try_apply : ('a -> 'a option) -> 'a -> 'a`?

transform_cleanup.ml
```diff
@@ -0,0 +1,69 @@
...
+         * is excluded by the second tests "succ (pc-1) = [pc]")
+         * TODO: we should implement some generic api for instructions like, "Instr.is_jmp" *)
+      Remove 1
+    | i, _ ->
+      Unchanged
+  in
+  change_instrs transform instrs
+
+let remove_unreachable_code (instrs : instruction_stream) : instruction_stream option =
+  let unreachable =
```
gasche: I'm only seeing this now and it's unrelated to the current commit, but calling this `reached` or `reachable` makes more sense than `unreachable`.

instr.ml
```diff
@@ -137,6 +137,7 @@ type program = {
   main : afunction;
   functions : afunction list;
 }
+type analysis_input = ModedVarSet.t * instructions
```
gasche: You should either have a comment explaining what the first component is (a set of formal parameters), or make this a record: `{ formals : ModedVarSet.t; instrs : instructions }`.

sourir.mllib
```diff
@@ -8,10 +8,11 @@ Eval
...
-Hoist_assign
-Cleanup
-Assumption
+Transform_hoist
+Transform_hoist_assign
+Transform_utils
+Transform_constantfold
+Transform_liveness
+Transform_cleanup
+Transform_assumption
```
gasche, line 18: Wouldn't it make sense to rebase this file change into the previous commit?

transform.ml
```diff
@@ -1,7 +1,7 @@
 open Instr
 
 (* types of transformations *)
-type transform_instructions = instruction_stream -> instruction_stream option
+type transform_instructions = analysis_input -> analysis_input option
```
gasche: Are there concrete cases where we want to remove some of the formal parameters? If not, would it be better to return `instructions`?

transform_hoist_assign.ml
```diff
@@ -13,12 +13,13 @@ open Instr
...
  * variable to avoid overriding unrelated uses of the same name.
  *)
-let hoist_assignment (instrs : instruction_stream) : instruction_stream option =
-  let reaching = Analysis.reaching instrs in
-  let uses = Analysis.used instrs in
-  let dominates = Analysis.dominates instrs in
+let hoist_assignment ((formals, instrs) as inp : analysis_input) : analysis_input option =
+  let open Analysis in
+  let reaching = reaching inp in
+  let uses = used inp in
```
gasche: the name used on both sign of the equal sign is different here, and the same on the two other cases. Which of `uses` or `used` is better?

gasche (commented):
The work in this PR introduces some complexity in the implementation. It is a good example of the idea that introducing aliasing makes things harder (or that the "big-step" approach to optimization keeps growing into a more and more elaborate system as the language gains weight), but I am somewhat melancholic of the previous simpler setting. Oh well.

sourir.ml
o-: an there you are saying you want less complexity :) with this pr we just have one option, I think a framework is overblown, not?

transform.ml
o-: no, but the reason was that it allows one to combine transformations with &&&.
but now that I think about it, the &&& operator could take care of pushing it through!

o-:
hey, thanks for your comments. I'll deal with them tomorrow. I must agree that I am sad how complex the thing became. Though I don't think that this PR adds to it. First it reduces some because the api is now uniform everything around the transformations we only have to deal once. Secondly it was a lie in the first place because we had mutable variables but no way of creating an alias. Thus we treated them like local variables all over the place. In the end I think it would have been less of a headache if we added functions from the beginning...

sourir.ml
gasche: The "framework" gives you shorter code and `--help` for free, and makes it easy to extend with other options in the future. I think that `Cmdliner` may be overblow but `Arg` is not.

transform_hoist_assign.ml
mhyee: `uses` would be consistent with `dominates` (and also the comment). But then we also have `required`, `required_vars`, and `aliased`. (I'm referring to the definitions in `analysis.ml`.)

At the very least, in this part of the code `uses` is consistent with how the variable is used (e.g. `dominates_all_uses`).

transform_hoist.ml
```diff
@@ -188,6 +158,21 @@ let pull is_target cond instrs =
 
 
 module Drop = struct
+  (* TODO:
+   * If we ever add mutable return values or suplicating refs this breaks!
```
mhyee: Did you mean **duplicating**?

transform_hoist_assign.ml
```diff
@@ -0,0 +1,55 @@
...
+    PcSet.for_all (fun use -> dominates pc use) (uses pc) in
+  let aliased = aliased inp in
+  let aliased var pc = VarSet.mem var (aliased pc) in
+  let rec find_possible_move pc =
+    if pc = Array.length instrs then None
+    else
+      let pc' = pc + 1 in
+      match[@warning "-4"] instrs.(pc) with
+      | Assign (x, exp) ->
+        if not (dominates_all_uses pc) || (aliased x pc) then find_possible_move pc'
```
mhyee: I don't have the operator precedences memorized, so I'm really paranoid about conditions like this (especially when negation is thrown in).

edit.ml
```diff
@@ -44,3 +77,122 @@ let subst_many instrs substs =
...
+  | Assign (x, exp) ->
+    let fresh = fresh_var instrs x in
+    instrs.(def) <- Decl_mut (fresh, Some exp);
+    List.iter (fun pc ->
+      instrs.(pc) <- replace_uses_in_instruction x fresh instrs.(pc)) uses
+  | _ ->
+    assert(false)
+open Instr
+
+let var_in_simple_exp var (exp : simple_expression) (in_exp : simple_expression) : simple_expression =
```
mhyee: So this code was originally from `replace.ml`, and there's very similar code above from `rename.ml`. Replacing/renaming seems to be very similar, but maybe there's a reason why we can't have them share code?

But more importantly, I think these functions need to be renamed. Before, `Rename.in_simple_expression` and `Replace.var_in_simple_exp` were distinct. But now, `Edit.var_in_simple_exp` isn't clear. The names are also a little inconsistent.

sourir.ml
mhyee: Skimming the documentation, `Arg` looks like it would be appropriate here. At the very least, it would get rid of the ugly substring manipulation.

I'll open an issue, since this PR is already really big and the current implementation is OK for now. It's also a lower priority task and could be a nice intro to OCaml for someone joining the project.

transform.ml
mhyee: I second this. Seeing `!!`, `!!!`, and `&&&` for the first time was a bit confusing.

transform.ml
o-: Ah so the problem is that if transform_instructions is not transitive I cannot use the same operator for chaining instruction transformations and program and function transformations.

I have changed it now (because I agree that outputting the formals is dumb). To fix the operator problem I have introduced two new stream operators. (==>) makes a transform_instruction transitive and (<==) turns it back into a normal transform_instruction. Now to chain instruction transformation you have to:

`(<==) ((==>) t1 &&& (==>) t2)`

mhyee:
Sorry, but I'm going to push back against the operators ( `&&&`, `==>` and `<==`, since `!!` and `!!!` are now removed). I didn't really like `&&&`; without looking at its definition, I have no idea what it does, or that it even executes the transformations right-to-left. (And grepping the codebase for `&&&` is annoying.)

But now you have to write:

```ocaml
let cleanup_all_instrs = (<==)
  ((==>) Transform_cleanup.remove_jmp &&&
   ((==>) Transform_cleanup.remove_unused_label &&&
    ((==>) Transform_cleanup.remove_unused_decl &&&
     (==>) Transform_cleanup.remove_unreachable_code)))
```
Which is even more opaque.

Could you rename `&&&`? Would it be possible/clearer to have something like `t1 <then> t2`?

But that doesn't solve `<==` or `==>`. Maybe a `combineN t1 t2 t3 ... tN` function (for `N` up to 5) where `<==` and `==>` are only used within `combineN` (and `<==` and `==>` are defined right before `combineN`, the point is to keep its definition next to its use)? A `combine` that takes a list of functions?

Or maybe the lesser evil is to keep passing the formals around.

(I'll look at the other changes later, I have to run right now.)

o-:
> Could you rename &&&? Would it be possible/clearer to have something like t1 <then> t2?

I don't know how to make infix operators with a string. It's supposed to be just the normal math function composition operator. Maybe we can take `%` apparently that is used for function composition in the batteries library.

> Maybe a combineN t1 t2 t3 ... tN

I don't like that proposal. It's too verbose

> Or maybe the lesser evil is to keep passing the formals around.

maybe it is after all

gasche:
> I don't like that proposal. It's too verbose

One option would be `instr_transforms [t1; t2; t3; ...; tN]`. It's not actually very verbose -- and keep in mind that we expect these operators to only appear a handful of times in the code, to describe high-level combination of passes.

transform.ml
```diff
@@ -1,205 +1,124 @@
...
+        instrs;
+        annotations = None } in
+      Some (Instr.replace_active_version func res)
 
-          begin match Analysis.find_first instrs valid_move with
-          | exception Not_found -> find_possible_move pc'
-          | pc' -> Some (pc, pc')
-          end
+let as_opt_program (transform : opt_function) : opt_prog =
+  let transform instrs =
```
mhyee: Could you rename this? It shadows the `transform` parameter, and I was confused why on line 56, `transform prog.main` was returning two values.

mhyee:
Right-to-left (as in function composition) makes total sense when you're writing something like `(pass2 &&& pass1) instrs`. But it's a bit strange when you're writing `let combined = pass2 &&& pass1; combined instrs`.

I guess it's not possible to have infix functions, the way you would with backticks in Haskell. I think I'd prefer the more verbose `instr_transforms [t1; t2; ...; tN]` over using an operator (`&&&` or `%`), because it's clearer.

o-:
ok, I am convinced. @mhyee you are right it was ugly :)

this is how it looks now:
https://github.com/reactorlabs/sourir/pull/87/commits/054460c06ddeb75e33591d0518ae3cfff4e5910b#diff-e87230a5806e9927be124ca6f5a2893cR74

(please forgive for not backporting it one commit earlier because then I'd have to do it twice since the signatures changed...)

and thanks for the list proposal. It's great

transform.ml
o-: I guess I can. Though I must say I this was on purpose and I find it quite legitimate. I want to specialize the argument to the local use case so I override it with the tailored version. Saves me from using the wrong one...

mhyee: Oh, you already made the change. I was going to retract my comment (as long as you commented the tailored version), because it was also my fault for not reading carefully enough. Sorry to make you change it and then revert.

mhyee:
Combining transformations looks a lot better now, thanks!

transform.ml
```diff
@@ -38,33 +42,45 @@ let as_opt_function (transformation : transform_instructions) : opt_function =
...
-  (Transform_cleanup.remove_jmp &&&
-   (Transform_cleanup.remove_unused_label &&&
-    (Transform_cleanup.remove_unused_decl &&&
-     Transform_cleanup.remove_unreachable_code)))
-let minimize_liverange_instrs = Transform_liveness.minimize_liverange &&& Transform_cleanup.remove_unused_decl
-let const_prop_instrs = Transform_cleanup.remove_unused_decl &&& Transform_constantfold.const_prop
+let cleanup_all_instrs = combine_transform_instructions [
+    Transform_cleanup.remove_unreachable_code;
+    Transform_cleanup.remove_unused_decl;
+    Transform_cleanup.remove_jmp;]
```
gasche: super-nitpicky (as in: no need to change): I prefer to have `]` on its own line because it makes nicer diff when we insert a new line at the end. (It's not just taste, optimizing for diff readability is an *objective* criterion.)

gasche:
I'm happy with the return over investment in the bike-shedding session, so I would be happy to get the PR merged now.

transform.ml
o-: good point. I need to remember it when I argue with lispers

transform.ml
o-: don't worry. I don't mind how it is now. It seems we have a different taste regarding shadowing. I overuse and you probable underuse it :)

o-:
super. @mhyee happy too?

transform.ml
mhyee: Let's just take the average then.

mhyee:
Yes.

PR #89, opened 2017-04-12
Assumptions

o-:
This PR abandons the good old branch pruning and replaces it with a more generalized version.

The first commit adds multiple osr conditions to a single osr instruction. The new syntax is `osr [condition, ...] (target) [osr-map]`. As an example `osr [(a==2), (b==4), (a<>b)] (main,old,l3) [...]`. It triggers if any of the conditions evaluates to true.

Then the big picture of the approach is to:

1. Insert checkpoints at every point in the program. Checkpoints are valid osr with empty condition list and a matching label.
2. Insert assumptions at will (ie. add an osr condition to an existing checkpoint). Assumptions can be thought of as quasi invariants we want to ensure at a certain location in the program.
3. Use the assumptions for optimizations. At location x, as long as all intermediate instructions (on all possible traces) between a guarding osr and x do not interfere with an assumption it can be assumed to hold at x.
4. Remove superfluous (empty) checkpoints.

Steps 1 and 4 are universal. For step 2 many ways of coming up with good assumptions are conceivable. Step 3 applies to any classical compiler optimization we have.

To elaborate a bit more on checkpoints. Assuming the following program:
```
version base
  mut x = 1
  return x
```
This is how it looks after checkpoint insertion:
```
version base
    mut x = 1
  checkpoint_1:
    osr [] (base, checkpoint_1) [mut x = &x]
    return x
```
To add assumptions we create new versions and let the targets point one version down
```
version base_2
    mut x = 1
  checkpoint_1:
    osr [assumption1, assumption2] (base_1, checkpoint_1) [mut x = &x]
    return x
version base_1
    mut x = 1
  checkpoint_1:
    osr [assumption1] (base, checkpoint_1) [mut x = &x]
    return x
version base
    mut x = 1
  checkpoint_1:
    osr [] (base, checkpoint_1) [mut x = &x]
    return x
```
Future work would be to merge multiple versions into two (most optimized and base). This is trivial and could be done on the fly. But conceptually I like the gradual approach (it also means you do not have to undo all assumptions at once).

The strategy for combined compiler optimization with assumption insertion is

1. Insert checkpoints
2. Run optimizers
3. Run passes which insert assumptions
4. Hoist assumptions (see below)
5. Repeat 2 - 4
6. Remove empty checkpoints
7. Run optimizers (repeat)

Step 7 is important since empty checkpoints keep potentially more variables alive than necessary (blocking optimizations). Note that theoretically after step 6 it is still possible to add new assumption. But it is not guaranteed to succeed because no fitting checkpoint might be available to insert them.

The branch_pruning is rewritten to fit this strategy. It is now performed in two passes. First a pass which inserts assumptions matching the branch conditions. And second simple optimization which turns `branch e l1 l2` into `goto l2` if e can be shown false.

Additionally, the 3rd patch introduces an optimization that hoists assumptions. If we have two osr and the intermediate instructions do not interfere we can move conditions from one to the other:
```
osr [(a=1)]
<some-instrs>
osr [(a=2)]
====>
osr [(a=1), (a=2)]
<some-instrs>
osr []
```
If in between there is a label and all predecessors except the dominating one have the assumption available, we can hoist further:
```
1   osr [(a=1)]
2  loop:           # One predecessor is dominating. The other one has the condition available.
3   osr [(a=2)]
4   <some-instrs>
5   goto loop      # (a=2) is available here if <some-instrs> do not interfere with it
====>
 osr [(a=1), (a=2)]
loop:
 osr []
 <some-instrs>
 goto loop
```

As an example of how powerful this new approach is consider the following optimization

```
 mut x
 mut y = 1
 read x
 mut i = 0
loop:
  branch (x==1) c1 c2
 c1:
  y <- 2
 c2:
  i <- (i+y)
 branch (i==10) end loop
end:
```

is optimized to

```
 mut x
 read x
 osr [(x == 1)] (main, anon_1, checkpoint_3) [mut x = &x, mut y = 1]
 mut i = 0
loop:
 i <- (i + 1)
 osr [(i == 10)] (main, anon_1, checkpoint_10) [mut i = &i, mut x = &x, mut y = 1]
 goto loop
```

### DIFF
--- a/analysis.ml
+++ b/analysis.ml
@@ -321,3 +321,26 @@ let aliased ({formals; instrs} : analysis_input) : pc -> VarSet.t =
   in
   let mut_formals = List.fold_left ref_param [] (ModedVarSet.elements formals) in
   fun _ -> VarSet.of_list mut_formals
+
+module Expression = struct
+  type t = expression
+  let compare (x : expression) (y : expression) =
+    Pervasives.compare x y
+end
+
+module ExpressionSet = Set.Make(Expression)
+
+let valid_assumptions {instrs} : pc -> ExpressionSet.t =
+  let merge _pc cur incom =
+    let merged = ExpressionSet.inter cur incom in
+    if ExpressionSet.equal merged cur then None else Some merged
+  in
+  let update pc cur =
+    match[@warning "-4"] instrs.(pc) with
+    | Osr {cond} ->
+      List.fold_right ExpressionSet.add cond cur
+    | _ ->
+      ExpressionSet.filter (fun exp ->
+          Instr.independent instrs.(pc) exp) cur
+  in
+  forward_analysis ExpressionSet.empty instrs merge update
--- a/edit.ml
+++ b/edit.ml
@@ -158,9 +158,9 @@ let replace_uses_in_instruction old_name new_name instr : instruction =
   | Branch (exp, l1, l2) ->
     Branch (in_expression exp, l1, l2)
   | Osr {cond; target; map} ->
-    Osr {cond = List.map in_expression cond;
-         target;
-         map = List.map in_osr map}
+    let cond = List.map in_expression cond in
+    let map = List.map in_osr map in
+    Osr {cond; target; map}
   | Decl_mut (x, None)
   | Read x ->
     assert (x != old_name);
--- a/edit.ml
+++ b/edit.ml
@@ -157,10 +157,10 @@ let replace_uses_in_instruction old_name new_name instr : instruction =
     Print (in_expression exp)
   | Branch (exp, l1, l2) ->
     Branch (in_expression exp, l1, l2)
-  | Osr (exp, f, v, l, osrs) ->
-    Osr (in_expression exp, f, v, l,
-         List.map in_osr osrs)
-
+  | Osr {cond; target; map} ->
+    Osr {cond = List.map in_expression cond;
+         target;
+         map = List.map in_osr map}
   | Decl_mut (x, None)
   | Read x ->
     assert (x != old_name);
--- a/eval.ml
+++ b/eval.ml
@@ -337,32 +337,32 @@ let reduce conf =
       pc = pc';
     }
   | Print e ->
-     let v = eval conf e in
-     { conf with
-       trace = v :: conf.trace;
-       pc = pc';
-     }
-  | Osr (e, f, v, l, osr_def) ->
-     let b = get_bool (eval conf e) in
-     if not b then
-       { conf with
-         pc = pc';
-       }
-     else begin
-       let osr_env, heap' = build_osr_frame osr_def conf.env conf.heap in
-       let osr_func = Instr.lookup_fun conf.program f in
-       let osr_version = Instr.get_version osr_func v in
-       let osr_instrs = osr_version.instrs in
-       { conf with
-         pc = resolve osr_instrs l;
-         env = osr_env;
-         heap = heap';
-         instrs = osr_instrs;
-         cur_fun = osr_func.name;
-         cur_vers = osr_version.label;
-         deopt = Some l;
-       }
-     end
+    let v = eval conf e in
+    { conf with
+      trace = v :: conf.trace;
+      pc = pc';
+    }
+  | Osr {cond; target={func;version; label}; map} ->
+    let triggered = List.exists (fun cond -> get_bool (eval conf cond)) cond in
+    if not triggered then
+      { conf with
+        pc = pc';
+      }
+    else begin
+      let osr_env, heap' = build_osr_frame map conf.env conf.heap in
+      let func = Instr.lookup_fun conf.program func in
+      let version = Instr.get_version func version in
+      let instrs = version.instrs in
+      { conf with
+        pc = resolve instrs label;
+        env = osr_env;
+        heap = heap';
+        instrs = instrs;
+        cur_fun = func.name;
+        cur_vers = version.label;
+        deopt = Some label;
+      }
+    end
 
 let start program input pc : configuration = {
   input;
--- a/instr.ml
+++ b/instr.ml
@@ -325,6 +325,24 @@ let used_vars = function
         | Osr_mut_undef _ -> Simple (Constant Nil)) map in
     VarSet.union (list_vars cond) (list_vars exps)
 
+let changed_vars = function
+  | Call (x, _, _)
+  | Decl_const (x, _) -> ModedVarSet.singleton (Const_var, x)
+  | Decl_mut (x, Some _)
+  | Assign (x ,_)
+  | Drop x
+  | Clear x
+  | Decl_mut (x, None)
+  | Read x -> ModedVarSet.singleton (Mut_var, x)
+  | Return _
+  | Branch _
+  | Label _
+  | Goto _
+  | Comment _
+  | Print _
+  | Osr _
+  | Stop _ -> ModedVarSet.empty
+
 exception FunctionDoesNotExist of identifier
 
 let lookup_fun (prog : program) (f : identifier) : afunction =
@@ -347,3 +365,13 @@ module Value = struct
   let bool b : value = Bool b
 end
 
+let checkpoint_prefix = "checkpoint_"
+let is_checkpoint_label l =
+  let len = String.length checkpoint_prefix in
+  String.length l > len && (String.sub l 0 11) = checkpoint_prefix
+let checkpoint_label pc =
+  checkpoint_prefix ^ (string_of_int pc)
+
+let independent instr exp =
+  let changed = ModedVarSet.untyped (changed_vars instr) in
+  VarSet.is_empty (VarSet.inter changed (expr_vars exp))
--- a/instr.ml
+++ b/instr.ml
@@ -69,6 +69,8 @@ end
 
 type pc = Pc.t
 
+type unique_pos = {func : label; version : label; label : label;}
+
 type instructions = instruction array
 and instruction =
   | Decl_const of variable * expression
@@ -84,8 +86,8 @@ and instruction =
   | Label of label
   | Goto of label
   | Print of expression
-  | Osr of expression * label * label * label * osr_def list
   | Stop of expression
+  | Osr of {cond : expression list; target : unique_pos; map : osr_def list; }
   | Comment of string
 and osr_def =
   | Osr_const of variable * expression
@@ -182,6 +184,10 @@ let arg_vars = function
   | Arg_by_val e -> expr_vars e
   | Arg_by_ref x -> VarSet.singleton x
 
+let list_vars ls =
+  let vs = List.map expr_vars ls in
+  List.fold_left VarSet.union VarSet.empty vs
+
 let declared_vars = function
   | Call (x, _, _)
   | Decl_const (x, _) -> ModedVarSet.singleton (Const_var, x)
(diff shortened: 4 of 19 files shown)


o-:
This is getting better :)

Together with #91 I can now optimize

```
 mut x
 mut y = 1
 read x
 mut i = 0
loop:
  branch (x==1) c1 c2
 c1:
  y <- 2
 c2:
  i <- (i+y)
 branch (i==10) end loop
end:
```

to

```
 mut x
 read x
 osr [(x == 1)] (main, anon_1, checkpoint_3) [mut x = x, mut y = 1]
 mut i = 0
loop:
 i <- (i + 1)
 osr [(i == 10)] (main, anon_1, checkpoint_10) [mut i = i, mut x = x, mut y = 1]
 goto loop
```

o-:
In case you are happy with #91 this can be reviewed as well

o-:
I updated the comment above, added some explanations and rebased the PR. ptal

instr.ml
```diff
@@ -339,3 +366,13 @@ module Value = struct
...
   let bool b : value = Bool b
 end
+
+let maybe_osr_target l =
+  String.length l > 11 && (String.sub l 0 11) = "checkpoint_"
+
+let osr_target pc =
+  "checkpoint_" ^ (string_of_int pc)
+
+let independant instr exp =
```
mhyee: Typo: independ**e**nt

transform_assumption.ml
```diff
@@ -1,37 +1,192 @@
...
+  Some { func with
+         body = [
+           { label = version.label;
+             instrs = (!!) baseline;
+             annotations = None;
+             has_checkpoints = true } ] }
+
+(* Removes empty checkpoints.
+ * Note: this should only be done to the most optimized version, since this one
+ * is guaranteed to not be a osr target. If we wanted to remove unnecessary
```
mhyee: Typo: a**n** osr target

tests.ml
```diff
@@ -285,20 +285,20 @@ c:
  clear r
 "
 
-let test_branch_pruned =
-"function main ()
-version anon_1
- mut x = 9
+let test_branch_pruned = " mut x = 9
```
mhyee: Nitpick: Can the string be moved to the next line? For a moment, I almost missed the `mut x = 9`, and was confused at the `mut x = &x` below.

mhyee:
The code looks reasonable, but I'm not sure I entirely understand this proposal, so I'd like to discuss it at the next meeting.

o-:
> The code looks reasonable, but I'm not sure I entirely understand this proposal, so I'd like to discuss it at the next meeting.

see 1a935ecde3320f953049b50006595c4785ca7aa1

analysis.ml
```diff
@@ -320,3 +320,28 @@ let aliased ({formals; instrs} : analysis_input) : pc -> VarSet.t =
   in
   let mut_formals = List.fold_left ref_param [] (ModedVarSet.elements formals) in
   fun _ -> VarSet.of_list mut_formals
+
+module Expression = struct
+  type t = expression
+  let compare (x : expression) (y : expression) =
+    Pervasives.compare
+      (Disasm.disassemble_exp_s x)
+      (Disasm.disassemble_exp_s y)
```
gasche: Doesn't `Pervasives.compare x y` work just as well? Are there expressions that have different representations but that you want to consider equal?

eval.ml
```diff
@@ -286,32 +286,32 @@ let reduce conf =
...
-         deopt = Some l;
-       }
-     end
+    let v = eval conf e in
+    { conf with
+      trace = v :: conf.trace;
+      pc = pc';
+    }
+  | Osr {cond; target={func;version; label}; map} ->
+    let not_triggered = List.for_all (fun cond -> not (get_bool (eval conf cond))) cond in
```
gasche: It would be marginally more elegant to define `let triggered` with `List.exists`.

instr.ml
```diff
@@ -226,14 +229,17 @@ let required_vars = function
...
         | Osr_mut (_, e) -> e
         | Osr_mut_ref (_, x) -> Simple (Var x)
-        | Osr_mut_undef _ -> Simple (Constant Nil)) osr in
-    let exps_vars = List.map expr_vars exps in
-    List.fold_left VarSet.union (expr_vars e) exps_vars
+        | Osr_mut_undef _ -> Simple (Constant Nil)) map in
+    let vs1 = List.map expr_vars cond in
+    let vs2 = List.map expr_vars exps in
+    let v1 = List.fold_left VarSet.union VarSet.empty vs1 in
+    let v2 = List.fold_left VarSet.union VarSet.empty vs2 in
```
gasche: This and the `Call` case above are calling for of a `list_expr_vars` auxiliary function.

instr.ml
```diff
@@ -339,3 +366,13 @@ module Value = struct
   let int n : value = Int n
   let bool b : value = Bool b
 end
+
+let maybe_osr_target l =
+  String.length l > 11 && (String.sub l 0 11) = "checkpoint_"
```
gasche: Hardocded constants are bad ;-)

```ocaml
let prefix = "checkpoint_" in
let prefix_len = String.length prefix in
String.length l >= prefix_len && String.sub l 0 prefix_len = prefix
```

transform_assumption.ml
```diff
@@ -1,37 +1,192 @@
...
+        Insert [Label (osr_target pc); Osr {cond=[]; target; map=osr};]
+    in
+    if pc = 0 then Unchanged else
+    match[@warning "-4"] instrs.(pc) with
+    | Stop _ | Return _ | Label _ | Comment _ -> Unchanged
+    | Osr _ -> assert false
+    | _ -> create_checkpoint pc
+  in
+  let baseline = change_instrs transform inp in
+  let (!!) = function | Some i -> i | None -> instrs in
```
gasche: Cool trick:

```ocaml
let (|?) opt def = match opt with Some v -> v | None -> def in
```

transform_assumption.ml
```diff
@@ -1,37 +1,192 @@
...
+  | Some pc ->
+    begin match[@warning "-4"] instrs.(pc) with
+    | Osr {cond; target; map} ->
+      instrs.(pc) <- Osr {cond=osr_cond::cond; target; map};
+      Some { version with instrs }
+    | _ -> assert (false)
+    end
+  end
+
+let hoist_assumption ({instrs} as inp) : instructions option =
```
gasche: Maybe a comment that describes what this (non-trivial) function does from a high level could be helpful.

gasche (commented):
If find that `has_checkpoints` is weird. What is the invariant that is supposed to correspond to this value being `true`? Why aren't the further optimizations also valid on a version that does not have it `true`? Creating non-local dependencies between optimization passes in this way would be best avoided, as it also complicates reasoning. I wonder if a different way to formulate the same transformations could get us rid of them.

(One thing that I can see is that it is obviously wrong to insert assumptions without duplicating the version first -- `next_version`and `insert_assumption` are linked together -- but that is not what the boolean currently check. I would find it a bit nicer to just merge the two transformations together, in the sense that `insert_assumption` would call `next_version` itself of having it as a precondition. But in any case the boolean, if it exists at all, should really correspond to the necessary-for-correctness part,.

Separately: even if we cannot remove labels from jumped-to versions, we can remove the `osr` instruction. This suggests splitting the cleanup in two. (Removing labels is indeed a global transform if anyone can jump to any label).

I haven't finished reviewing the code (it's a lot!), I stopped at `hoist_assumption` where the serious stuff begins.

o-:
On 04/25/2017 08:54 PM, Gabriel Scherer wrote:
> If find that `has_checkpoints` is weird.

yeah I think I'm gonna remove it.

> I haven't finished reviewing the code (it's a lot!), I stopped at `hoist_assumption` where the serious stuff begins.

is it a good time to fix what you said so far?

do you want me to split the hoisting into a different PR?

gasche:
Any time is good -- we can iteratively work on a PR.

o-:
ok, thanks for the comments so far. I have tried to address. ptal

gasche:
I'm afraid I created conflicts by merging #105. Would you rebase the PR when you have time? I'm looking at it again now in any case.

edit.ml
```diff
@@ -154,10 +154,10 @@ let replace_uses_in_instruction old_name new_name instr : instruction =
...
   | Branch (exp, l1, l2) ->
     Branch (in_expression exp, l1, l2)
-  | Osr (exp, f, v, l, osrs) ->
-    Osr (in_expression exp, f, v, l,
-         List.map in_osr osrs)
-
+  | Osr {cond; target; map} ->
+    Osr {cond = List.map in_expression cond;
+         target;
+         map = List.map in_osr map}
```
gasche: Nitpick:

```ocaml
let cond = List.map in_expression cond in
let map = List.map in_osr map in
Osr {cond; target; map}
```

transform_assumption.ml
```diff
@@ -1,37 +1,239 @@
...
-let result_version (func : afunction) (label : label) (instrs : instructions) : version =
-  { label = Edit.fresh_version_label func label;
-    instrs;
-    annotations = None }
+(* Inserts checkpoints at all program points
+ * A checkpoint is a well-formed osr point and a matching label.
+ * For example:
+ *   checkpoint_1:
+ *     osr [] (func, version, checkpoint_1) [mut x = &x]
+ * Initially the checkpoint is self referential (ie. its target is
```
gasche, line 9: The self-referentiality is tricky (if there was a condition that could fail here, we would get an infinite loop), but I don't see a better way to do this.

transform_assumption.ml
```diff
@@ -1,37 +1,239 @@
...
+  let transform pc =
+    match[@warning "-4"] instrs.(pc) with
+    | Osr {cond} when cond = [] -> Remove 1
+    | _ -> Unchanged
+  in
+  change_instrs transform inp
+
+(* Removes all checkpoint labels. Can be applied as a final cleanup
+ * when there are no osr-in points. *)
+let remove_checkpoint_labels ({instrs} as inp) : instructions option =
```
gasche: This optimization feels wrong to me as it relies on the implicit knowledge that there is no other version that points to these checkpoints. I would rather have a whole-program remove-unused-label optimization.

gasche (approved):
Thanks a lot for the extra comments which were super-helpful. I now understand the code and I think we should merge in -- as soon as it is conflict-free -- to avoid conflictrot.

I think that we could make the codebase better by improving the (already pretty cool) `valid_assumptions` function. Currently it only does forward propagation of assumptions, and the hoisting logic does backward propagation. We could have a `valid_assumptions` function that would saturate in both forward and backward directions, and then the `hoist_assumption` logic could be sensibly simpler: instead of one simple-block and one out-of-loop function, we could have a single pass that goes backward, duplicating the to-be-inserted osr on multi-predecessor points, and stopping as soon as one of the predecessors does not validate the assumption.

We could implement this as a further PR.

transform_assumption.ml
o-: agreed. I create an issue for it (I'd rather not completely disable it because it makes the output of the optimizer super annoying to read...)

o-:
So I merge it as is and create issues for the remaining things. I hope that's ok. Also I removed the disassembler refactoring commit because it's to entangled with 8577c8c5db0e08e04a4adc1d37d2b925af32053f

PR #90, opened 2017-04-13
fix drop hoisting across calls

o-:
On a call instruction

call x = ...

x is not a required variable. It still has to block because unlike
`mut x` or `const x` it's not annihilating the call.

### DIFF
--- a/tests.ml
+++ b/tests.ml
@@ -947,6 +947,14 @@ let do_test_push_drop () =
    l:
    " in
   test t 7 e;
+  let t = parse "
+      const a = 1
+      call x = bla ()
+      drop x
+    function bla ()
+      return 1
+   " in
+  test t 2 t;
   ()
 
 let do_test_drop_driver () =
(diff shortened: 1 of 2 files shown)


gasche:
I was about to propose a `call _ = ...` form to rewrite to in this case. But in fact I think it's not necessary because it is easy to convince ourselves that blocking hoisting here will not prevent further optimizations: because `call x` is defining `x`, we now that `x` is not in scope above, so moving the drop above would not unlock further optimization opportunities.

o-:
well moving the drop would be invalid because you try to drop a variable which is not in scope :)

was that a lgtm?

mhyee:
Looks good.

Minor: could you add a short comment explaining why `call` is not annihilating? (For a second, I forgot that we have mutable "out" params now.)

Tangential: Do we even want a `call _ = ...` form? On one hand, I don't really want to change too many things. On the other hand, it might make sense if a function *only* returns via out params, or a function is called only for side-effects (printing). But there's nothing wrong with a dummy variable, right?

(And now this got merged while I was writing. Whoops.)

o-:
> (For a second, I forgot that we have mutable "out" params now.)

That's not the reason. Call can have side-effects, you cannot just skip it, just because the result is not used.

Issue #48, opened 2017-02-08
osr variable declarations: separate rebinding and redeclaration


gasche:
In #45 we discussed the syntax of osr redeclaration zone, and I made a proposal that is a bit more complex/ambitious (but also more regular and, I think, a bit nicer) than what is currently supported:

> We could do a clear distinction between renamings and new declarations (in terms of the existing scope). We could for example have x <- y, which requires that both variables be of the same mutability class (what you call "type"), and then the general declaration forms const x = e, mut x = e and mut y, that are understood to declare new variables.

@o- answered with common sense and minimalism:

> yeah, this might be interesting in the future. But for all we need osr now it's very easy to get back to the non-allocating version by not having expressions for mutable variables.

If/when we have time (or someone wants to scratch and itch), it might be interesting to try to do the more complete version.

mhyee:
Note: the discussion is actually [here](https://github.com/reactorlabs/sourir/pull/46#discussion_r100071371), in #46.

mhyee:
Fixed by #91.

PR #91, opened 2017-04-13
implement a simple mut-to-const pass

### LINKED ISSUE #69: OsrMut should take an expr, not a var
https://github.com/reactorlabs/sourir/blob/master/instr.ml#L49

We want:
```ocaml
and osr_def =
  | OsrConst of variable * expression
  | OsrMut of variable * expression (* This was variable *)
  | OsrMutUndef of variable
```

There was probably a reason why it wasn't like this, but we should investigate.

o-:
to reduce the friction I changed osr to take an expression in the mut case. This resolves #69 en passant.

### DIFF
--- a/transform_constantfold.ml
+++ b/transform_constantfold.ml
@@ -45,14 +45,16 @@ let const_prop ({formals; instrs} : analysis_input) : instructions option =
       Assign (y, replace e)
     | Branch (e, l1, l2) -> Branch (replace e, l1, l2)
     | Print e -> Print (replace e)
-    | Osr (exp, f, v, l, env) ->
+    | Osr (exp, tf, tv, tl, env) ->
       (* Replace all expressions in the osr environment. *)
       let env' = List.map (fun osr_def ->
-        match[@warning "-4"] osr_def with
+        match osr_def with
         | Osr_const (y, e) -> Osr_const (y, replace e)
-        | _ -> osr_def) env
+        | Osr_mut (y, e) -> Osr_mut (y, replace e)
+        | Osr_mut_ref (y, z) -> if x=z then Osr_mut (y, Simple l) else osr_def
+        | Osr_mut_undef y -> Osr_mut_undef y) env
       in
-      Osr (replace exp, f, v, l, env')
+      Osr (replace exp, tf, tv, tl, env')
     | Drop y
     | Decl_mut (y, None)
     | Clear y
--- a/transform_constantfold.ml
+++ b/transform_constantfold.ml
@@ -106,3 +106,55 @@ let const_prop ({formals; instrs} : analysis_input) : instructions option =
   in
 
   work instrs
+
+open Transform_utils
+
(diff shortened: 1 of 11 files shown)


gasche:
I still dislike the idea that osr would always have to create a fresh mutable variable, instead of being able to alias an existing heap location. I think the change is a regression here -- see #45 and #48. We can have just `mut x = y` or have both `mut x <- y` and `mut x = e`, but having only the latter is meh.

o-:
good point. will do...

o-:
actually I don't think it's a good idea to have both `mut x = 1` and `mut x <- 1`.

Either we think osr should not change the heap in general (then we must have some other solution to the problem).

Or we accept that it does. In this case having the two osr cases just complicates matters for no good reason. I think whether the osr reuses a heap reference or not is an implementation detail.

In the function call case the distinction is relevant since one creates an alias and the other doesn't. But in this case the old reference dies. Therefore the two are equivalent. It's just an optimization that the osr mechanism can reuse the reference in one of the cases.

o-:
retraction. Writing the above comment made me realize that osr-by-value of an aliased mutable variable breaks the semantics! So we need the two cases

o-:
ptal

transform_constantfold.ml
```diff
@@ -104,3 +106,55 @@ let const_prop ({formals; instrs} : analysis_input) : instructions option =
...
+
+let make_constant (({formals; instrs} as inp) : analysis_input) : instructions option =
+  let required = Analysis.required inp in
+  let constant var pc =
+    match[@warning "-4"] instrs.(pc) with
+    | Assign (x, _) when x = var -> false
+    | Drop x when x = var -> false
+    | Clear x when x = var -> false
+    | Read x when x = var -> false
+    | Call (_, _,exp) ->
```
mhyee: Nitpick: missing space before `exp`.

mhyee:
Looks good to me.

gasche:
I'm sorry for the nitpick, but a minor thing is itching me with the merged version: I would rather expect `mut x <- y` and `mut x = e`, because the latter looks like the declaration syntax (so it makes sense for it to be general (accept expressions) and have the same semantics), while the former looks more like "x is replaced by y" to me, which (to me) suggests the reuse of the heap cell. Does someone have a reason of taste for the current choice which, I believe, is the opposite one, `mut x <- e` and `mut x = y`?

o-:
hmm I don't really have a good reason for picking one over the other now that I think about it. Honestly I think there lies no good answer here. The existing `<-` does not declare a variable, but updates it (also `a <- b` does not create an alias but assignes "by-value"). Which does not correspond to either of the osr cases. To be honest the most consistent with what we have would be `mut x = e` and `mut x = &y`. If you don't like the `&` we should find a new syntax for both the args and the osr, but I feel like they should be consistent.

o-:
let's continue at #92

PR #92, opened 2017-04-19
change osr-def syntax

o-:
use the same syntax for references as in the call case

Also adds some tests

### DIFF
--- a/doc/ir-semantics.text
+++ b/doc/ir-semantics.text
@@ -61,12 +61,13 @@ v :=     values
 osr-map ::=    osr mapping declaration
 | const x = e
 | mut x
-| mut x = x
+| mut x = &x
+| mut x = e
 
 # Natural operational semantics
 
 E ::= (x ↦ val v | x ↦ ref a | x ↦ ⊥)*  lexical environment
-H ::= (l ↦ v)                           mutable heap
+H ::= (a ↦ v)                           mutable heap
 
 T ::= (lit)*                    output trace
 
@@ -110,10 +111,18 @@ Update (partial) function, returns an H:
 
 ## Osr transformation of environment
 
-  osr-env P H E (const x = e) = (x ↦ eval P H E e)
-  osr-env P H E (mut x)       = (x ↦ ⊥)
-  osr-env P H E (mut x = y)   = (x ↦ a)
+  materialize P H E (mut x = e) = (x ↦ (a ↦ v))
+    where a - fresh
+      and v := eval P H E e
+  materialize P H E _           = ()
+
+  osr-env P H E M (const x = e) = (x ↦ val v)
+    where v := eval P H E e
+  osr-env P H E M (mut x)       = (x ↦ ⊥)
+  osr-env P H E M (mut x = &y)  = (x ↦ ref a)
     where a = E(y)
+  osr-env P H E M (mut x = e)   = (x ↦ ref e)
+    where (a ↦ _) = M(x)
 
 ## Argument evaluation
 
@@ -166,17 +175,20 @@ Update (partial) function, returns an H:
                if v = false then l₂
 
   step   osr(e, f^, lᵥ^, l^, (osr-map*)) (P, I, T, H, E, l)
-       = (I', T, H, E', l')
+       = (I', T, H', E', l')
     when v := eval P H E e
      and if v = false then
          I' := I
          l' := l
+         H' := H
          E' := E
      and if v = true then
          Df ∈ dom P
          Df  = Function f^ (formal*)
          I' := P(Df, lᵥ^)
          l' := l^
+         M  := (materialize P H E osr-map)*
+         H' := H ∪ rng(M)
          E' := (osr-env P H E osr-map)*
 
 # Scopes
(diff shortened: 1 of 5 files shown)


mhyee:
I honestly have no intuition/preference between `mut x <- y`/`mut x = e` or `mut x <- e`/`mut x = y` (which is probably a reason to change it). But I agree consistency with the call syntax is probably better.

For the tests, could you also add a case where we have a non-variable expression?

Also, don't forget to update the semantics doc.

o-:
> Also, don't forget to update the semantics doc.

riight. done

o-:
> For the tests, could you also add a case where we have a non-variable expression?

Don't understand. There is a test with const. I guess you meant something diffrent.

Update: sorry, get it

parser.mly
```diff
@@ -110,9 +110,9 @@ osr_def:
     { Osr_const (x, e) }
 | MUT x=variable
     { Osr_mut_undef x }
-| MUT x=variable EQUAL y=variable
+| MUT x=variable EQUAL AMPERSAND y=variable
     { Osr_mut_ref (x, y) }
-| MUT x=variable LEFTARROW e=expression
+| MUT x=variable EQUAL e=expression
```
gasche, line 115: Do we need to update the parser? I wonder if some error messages mention this syntax.

gasche (approved):
Ready for merge -- maybe see the parser.messages question.

parser.mly
o-: yeah, I think they are out of sync anyways :(

doc/ir-semantics.text
```diff
@@ -166,17 +175,20 @@ Update (partial) function, returns an H:
...
+         H' := H
          E' := E
      and if v = true then
          Df ∈ dom P
          Df  = Function f^ (formal*)
          I' := P(Df, lᵥ^)
          l' := l^
+         M  := (materialize P H E osr-map)*
+         H' := H ∪ rng(M)
          E' := (osr-env P H E osr-map)*
```
mhyee, line 192: `M` was added to `osr-env`, so this should be: `E' := (osr-env P H E M osr-map)*`

doc/ir-semantics.text
```diff
@@ -110,10 +111,18 @@ Update (partial) function, returns an H:
...
+  materialize P H E (mut x = e) = (x ↦ (a ↦ v))
+    where a - fresh
+      and v := eval P H E e
+  materialize P H E _           = ()
+
+  osr-env P H E M (const x = e) = (x ↦ val v)
+    where v := eval P H E e
+  osr-env P H E M (mut x)       = (x ↦ ⊥)
+  osr-env P H E M (mut x = &y)  = (x ↦ ref a)
     where a = E(y)
```
mhyee, line 123: Looks like the notation `E ∋ (y ↦ ref a)` is used more often.

doc/ir-semantics.text
```diff
@@ -110,10 +111,18 @@ Update (partial) function, returns an H:
...
+  osr-env P H E M (mut x = e)   = (x ↦ ref e)
```
mhyee, line 124: Does `e` need to be evaluated?

mhyee: Oh, I think I get it. `e` is evaluated by `materialize`.

But then it should be `(x ↦ ref a)`, right?

doc/ir-semantics.text
```diff
@@ -110,10 +111,18 @@ Update (partial) function, returns an H:
...
+    where (a ↦ _) = M(x)
```
mhyee, line 125: Likewise, I think it would be more consistent to write `M ∋ (x ↦ (a ↦ _))`

doc/ir-semantics.text
```diff
@@ -61,12 +61,13 @@ v :=     values
...
 | mut x
-| mut x = x
+| mut x = &x
+| mut x = e
 
 # Natural operational semantics
 
 E ::= (x ↦ val v | x ↦ ref a | x ↦ ⊥)*  lexical environment
-H ::= (l ↦ v)                           mutable heap
+H ::= (a ↦ v)                           mutable heap
```
mhyee, line 70: While I'm at it, nitpick at how "mutable heap" doesn't line up with "lexical environment."

doc/ir-semantics.text
```diff
@@ -166,17 +175,20 @@ Update (partial) function, returns an H:
...
          l' := l
+         H' := H
          E' := E
      and if v = true then
          Df ∈ dom P
          Df  = Function f^ (formal*)
          I' := P(Df, lᵥ^)
          l' := l^
+         M  := (materialize P H E osr-map)*
+         H' := H ∪ rng(M)
```
mhyee, line 191: It took me a while to realize that `rng` was range. I'd prefer it to be explicitly written out.

mhyee:
I had some more nitpicking/stupid questions. Sorry I couldn't get them in before the merge, I was in class.

gasche:
Sorry for merging too fast! If you know what you would prefer to see instead for each of these changes, feel free to do them yourself and send a followup/cleanup PR.

mhyee:
Sure, will do so.

o-:
> Sure, will do so.

I can fix it too, but if you want, go ahead

doc/ir-semantics.text
mhyee: Turns out the columns line up, but some symbols aren't the same width? Oh well.

PR #94, opened 2017-04-19
Fix nitpicks in ir-semantics.text

mhyee:
Fixes the late code review comments from #92.

### DIFF
--- a/doc/ir-semantics.text
+++ b/doc/ir-semantics.text
@@ -120,9 +120,9 @@ Update (partial) function, returns an H:
     where v := eval P H E e
   osr-env P H E M (mut x)       = (x ↦ ⊥)
   osr-env P H E M (mut x = &y)  = (x ↦ ref a)
-    where a = E(y)
-  osr-env P H E M (mut x = e)   = (x ↦ ref e)
-    where (a ↦ _) = M(x)
+    where E ∋ (y ↦ ref a)
+  osr-env P H E M (mut x = e)   = (x ↦ ref a)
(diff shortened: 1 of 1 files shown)


doc/ir-semantics.text
```diff
@@ -120,9 +120,9 @@ Update (partial) function, returns an H:
     where v := eval P H E e
   osr-env P H E M (mut x)       = (x ↦ ⊥)
   osr-env P H E M (mut x = &y)  = (x ↦ ref a)
-    where a = E(y)
-  osr-env P H E M (mut x = e)   = (x ↦ ref e)
-    where (a ↦ _) = M(x)
+    where E ∋ (y ↦ ref a)
+  osr-env P H E M (mut x = e)   = (x ↦ ref a)
+    where M ∋ (x ↦ (a ↦ _))
```
gasche, line 125: I think this line should evaluate `e` first and put the result in the memory at `a`.

gasche (changes requested):

doc/ir-semantics.text
mhyee: That's what I thought at first, but `materialize` does the evaluation of `e` and updates the heap. So the result is already in `M` before `osr-env` is called. (And `osr-env` only updates the environment.)

On one hand, it makes sense to keep the two separate, On the other hand, it might be clearer to update both the heap and environment in the same place.

gasche: Urgh, ok. I'll update to update both at once.

gasche (approved):

PR #95, opened 2017-04-19
[doc] factorize environment creation across declarations, calls and OSR

gasche:


doc/ir-semantics.text
```diff
@@ -111,18 +111,15 @@ Update (partial) function, returns an H:
...
-      and v := eval P H E e
-  materialize P H E _           = ()
-
-  osr-env P H E M (const x = e) = (x ↦ val v)
+  create-env H E (const x = e) = (∅, [x ↦ val v])
     where v := eval P H E e
-  osr-env P H E M (mut x)       = (x ↦ ⊥)
-  osr-env P H E M (mut x = &y)  = (x ↦ ref a)
+  create-env H E (mut x)       = ([a ↦ ⊥], [x ↦ ref a])
+    where a fresh
```
o-: So far we stored ⊥ in the env. See the definition of envs `val v | ref a | ⊥`. If you want to store it in the heap then we'd have to update the definition of `E` and potentially other places too. I lean towards keeping it in the env. The reason is that (1) it's not necessary to create a heap cell. And (2) it should never be allowed to read an undefined variable. For example with your changes the following program would not error on line 2 but I think it should.

```
  mut x
  call foo(&x)

function foo(mut x)
  return 0
```

Update: To be precise the checker would still complain. But at runtime it wouldn't error anymore.

doc/ir-semantics.text
```diff
@@ -134,13 +131,13 @@ Update (partial) function, returns an H:
 
   (P, C*) : (I, T, H, E, l) --> (P, C*) : step I(l) (P, I, T, H, E, l)
 
-  (P, C*) : (I, T, H, E, l) --> (P, (C*, C')) : (I', T, H, E', start)
+  (P, C*) : (I, T, H, E, l) --> (P, (C*, C')) : (I', T, H', E', start)
```
o-: should be `H+H'`

doc/ir-semantics.text
```diff
@@ -152,13 +149,12 @@ Update (partial) function, returns an H:
 ## One step reduction
 
   step   (const x = e) (P, I, T, H, E, l)
-       = (I, T, H, E[x ↦ v], succ I l)
-    when v := eval P H E e
+       = (I, T, H+H', E+E', succ I l)
+   where (H', E') := create-env H E (const x = e)
```
o-: Ugh, I much rather prefer duplication than reusing something in unrelated contexts. This feels a bit forced to me... (similar in spirit to inheriting from a completely unrelated superclass to acquire one useful method). To some extend I guess it's a matter of taste. But I may add it's brittle since the fact that declaration and argument creation can use the same api is somewhat accidental.

Also: H' is empty. At least write `where (0, E') := create-env`

doc/ir-semantics.text
```diff
@@ -187,9 +183,7 @@ Update (partial) function, returns an H:
          Df  = Function f^ (formal*)
          I' := P(Df, lᵥ^)
          l' := l^
-         M  := (materialize P H E osr-map)*
-         H' := H ∪ range(M)
-         E' := (osr-env P H E M osr-map)*
+         (H', E') := sum (create-env H E osr-map)*
```
o-: Again, on line 174 the out-arg needs to be `H+H'` not just `H'`. Would it be useful to use a different metavariable (like I did with M) to make it more apparent that `H'` is just a partial heap?

doc/ir-semantics.text
```diff
@@ -152,13 +149,12 @@ Update (partial) function, returns an H:
...
-    when v := eval P H E e
+       = (I, T, H+H', E+E', succ I l)
+   where (H', E') := create-env H E (const x = e)
 
   step   (mut x = e) (P, I, T, H, E, l)
-       = (I, T, H[a ↦ v], E[x ↦ a], succ I l)
-    when v := eval P H E e
-     and a - fresh
+       = (I, T, H+H', E+E', succ I l)
+   where (H', E') := create-env H E (mut x = e)
```
o-: dito

o- (commented):
Thanks, nice change. I like that we use the same mf for call and osr. But I think reusing it for decls is stretching it.

doc/ir-semantics.text
gasche: But that semantics means that assignment may allocate, right? (Doesn't that mean that the rule for assignment, or rather the `(H,E)[x <- v]` notation, is wrong as it doesn't handle the uninitalized case correctly?) I don't like this, I would prefer `mut x` to allocate rather than the first write, because it means the semantics is closer to the syntax -- and it makes the performance of assignment easier to reason about.

gasche:
Interesting. It is clear from @o-'s reply that this "minor refactoring" is not so minor and that I need to think harder about the semantics of these. Thanks for your comment, I'll try to look deeper into this.

doc/ir-semantics.text
o-: Oh, you are right. That was a lingering bug. I agree with your point. Lets go this way.

o-: Still we need to make some conscious decision on whether `&a` is valid on a reference which contains undefined.

mhyee: I think it should be invalid, especially since we can catch it statically. But does this introduce any other problems?

doc/ir-semantics.text
mhyee: I think that might help, but we also have partial environments (see line 157, `E+E'`).

`H+H'` looks weird to me. Updating the heap (and environment) directly would fix this, but then `create-env` gets uglier. Probably not worth it, especially when we have to call `create-env` for each `osr-map` or argument.

gasche:
I thought more about this. It's clear that we need to pass the whole heap no matter what (we could in fact do different, pass only the heap fragment that is needed for environment, and then merge the heap on function return, but that is less standard than the simple state-passing style for no benefits I can see). On the other hand, I don't really agree with @o-'s gut feeling that variable declarations and OSR/function calls should not be conflated. They both build or extend environments and I think it makes sense to share the logic (I tried to go further by proposing a named-argument syntax for function call). I'll try to redo a pass on the current proposal to distinguish the two cases more clearly, but keep the factorization.

gasche:
Hopefully the new commit provides a correct semantics and a cleaner distinction between the two sort of operations on the environment (create of a new environment from a previous one, or extension of an environment).

doc/ir-semantics.text
```diff
@@ -150,11 +158,11 @@ Update (partial) function, returns an H:
 
   step   (const x = e) (P, I, T, H, E, l)
        = (I, T, H+H', E+E', succ I l)
```
o-: `s/([HE])\+[HE]'/\1'/`

doc/ir-semantics.text
```diff
@@ -150,11 +158,11 @@ Update (partial) function, returns an H:
...
-   where (H', E') := create-env H E (const x = e)
+   where (H', E') := extend-env H E (const x = e)
 
   step   (mut x = e) (P, I, T, H, E, l)
        = (I, T, H+H', E+E', succ I l)
```
o-: dito

o- (commented):
> On the other hand, I don't really agree with @o-'s gut feeling that variable declarations and OSR/function calls should not be conflated. 

Disagreement is healthy. Except for the two typos it lgtm.

doc/ir-semantics.text
```diff
@@ -111,18 +111,23 @@ Update (partial) function, returns an H:
 
 ## Osr transformation of environment
```
mhyee: I guess this header is outdated, since the function is more general now.

doc/ir-semantics.text
```diff
@@ -111,18 +111,23 @@ Update (partial) function, returns an H:
...
+    where a fresh
+  build-env H E (mut x = e)   = ([a ↦ v], [x ↦ ref a])
+    where a fresh
+      and v := eval P H E e
+  build-env H E (mut x = &y)  = (∅, [x ↦ ref a])
     where E ∋ (y ↦ ref a)
-  osr-env P H E M (mut x = e)   = (x ↦ ref a)
-    where M ∋ (x ↦ (a ↦ _))
+
+  extend-env H E def := (H, E) + build-env H E def
```
mhyee: `def` was never really defined. Right now it's the same as `osr-map`. Should we just rename `osr-map` to `def` or something else?

doc/ir-semantics.text
```diff
@@ -111,18 +111,23 @@ Update (partial) function, returns an H:
...
+
+    (extends the lexical environment E)
```
mhyee: This is just a formatting nitpick, but does it look better to write:
```
Extends the lexical environment E:
    extend-env H E def [...]
```

mhyee:
For the record, I agree with @gasche's decision to combine osr/calls/defs. I think it looks much clearer now, thanks!

I had some minor comments (documentation), but other than that, looks good.

gasche:
I fixed the typos as suggested.

I think we should bite the bullet and allow `mut x = &y` as a declaration in the bytecode as well. Another thing is that we might need to understand whether `call x = ...` should also be part of osr-maps, or maybe revert on the `call/const` distinction. I think we could do this in a separate PR.

PR #97, opened 2017-04-28
Constant folding

mhyee:
Implements #73.

When reading the diff, it might be easier to ignore the deleted code in `Transform.constant_fold`, since I basically deleted everything and wrote constant folding from scratch.

I also updated the old constant propagation tests and added some new ones.

(One thing I noticed is that the other test drivers only optimize and compare the `main` function, ignoring everything else. This was because the tests were written before we had functions, and it's not a problem now because few of those tests involve functions. I fixed this for my constant propagation tests, but we might want to convert the other tests later.)

There's also a small fix: 95c8b75d43bcb41f9209f0ec5f4d67ae4ff27ab0. Due to the way the optimizer iterates over functions, the order in the optimized version would be reversed from the order in the original version. Not necessarily a bug, but it makes it easier to compare before/after optimizations.

### DIFF
--- a/transform_constantfold.ml
+++ b/transform_constantfold.ml
@@ -1,10 +1,10 @@
 open Instr
 
 (*
- * Constant folding
+ * Constant folding.
  *
  * This optimization combines constant propagation, copy propagation, and
- * constant folding.
+ * constant folding. We assume the program is well-scoped.
  *
  * The strategy is to maintain a "propagation environment" mapping variables to
  * simple expressions (constants or variables) and apply the `normalize`
@@ -20,36 +20,43 @@ let const_fold (({formals; instrs} as inp) : analysis_input) : instructions opti
 
   (* Perform constant and copy propagation on the expression `exp`, using
    * the mappings in `penv`.
-   * Returns an updated expression. *)
+   * Returns a pair of the updated expression and changed status. *)
   let propagate exp penv =
-    let try_prop x e =
+    let try_prop x (e, c) =
       if Prop_env.mem x penv
-      then Edit.replace_var_in_exp x (Prop_env.find x penv) e else e in
-    VarSet.fold try_prop (expr_vars exp) exp
+      then
+        (Edit.replace_var_in_exp x (Prop_env.find x penv) e, true)
+      else (e, c || false) in
+    VarSet.fold try_prop (expr_vars exp) (exp, false)
   in
 
   (* Constant fold the expression `exp`.
-   * Returns an updated expression. *)
+   * Returns a pair of the updated expression and changed status. *)
   let fold exp =
     match[@warning "-4"] exp with
-    | Simple _ -> exp
+    | Simple _ -> (exp, false)
     | Op (op, [Constant(v1); Constant(v2)]) ->
       begin match op with
-      | Eq -> Simple (Constant (Bool (Eval.value_eq v1 v2)))
-      | Neq -> Simple (Constant (Bool (Eval.value_neq v1 v2)))
-      | Plus -> Simple (Constant (Int (Eval.value_plus v1 v2)))
+      | Eq -> (Simple (Constant (Bool (Eval.value_eq v1 v2))), true)
+      | Neq -> (Simple (Constant (Bool (Eval.value_neq v1 v2))), true)
+      | Plus -> (Simple (Constant (Int (Eval.value_plus v1 v2))), true)
       end
-    | Op (_, _es) -> exp
+    | Op (_, _es) -> (exp, false)
   in
 
   (* Try to add a new mapping to the environment. `exp` is the simple expression
-   * being bound to the constant variable `x`. *)
-  let try_add x exp scope penv =
+   * being bound to the constant variable `x`.
+   * We need `pc` to infer the scope. *)
+  let try_add x exp pc penv =
     match exp with
     | Simple (Constant v) -> Prop_env.add x (Constant v) penv
     | Simple (Var v) ->
-      (* `exp` is a variable v, need to check if `v` is const or mut. *)
-      begin match scope with
+      (* We have the declaration `const x = v`. Check if `v` is const or mut.
+       * Normally, if `v` is const, at this point it would already be replaced
+       * by a constant value. However, if `v` is a const parameter, then we
+       * need to do the check here and copy propagate it. *)
+      let scope = Scope.infer inp in
+      begin match scope.(pc) with
       | DeadScope -> penv
       | Scope scope ->
         try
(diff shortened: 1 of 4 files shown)


transform_constantfold.ml
```diff
@@ -1,111 +1,130 @@
...
-          (* Already seen this pc, but continue searching the `rest` of the
-             worklist. *)
-          find_targets instrs x rest acc
-        else
-          (* Add the current `pc` to the accumulator and update the worklist
-             with our successors. *)
-          let succs = successors_at instrs pc in
-          find_targets instrs x (succs @ rest) (PcSet.add pc acc)
-      end
+    | [] -> Some instrs
```
o-: it would be nice to stick to the api and return None if no changes where made

transform_constantfold.ml
```diff
@@ -1,111 +1,130 @@
...
+    | (_, pc) :: rest when PcSet.mem pc seen -> work instrs rest seen
+    | (penv, pc) :: rest ->
+      let scope = Scope.infer inp in
```
o-: nitpick, you could do this only when you change the instrs somehow

transform_constantfold.ml
```diff
@@ -1,111 +1,130 @@
...
+  let try_add x exp scope penv =
+    match exp with
+    | Simple (Constant v) -> Prop_env.add x (Constant v) penv
+    | Simple (Var v) ->
+      (* `exp` is a variable v, need to check if `v` is const or mut. *)
+      begin match scope with
+      | DeadScope -> penv
+      | Scope scope ->
+        try
+          (* If `v` is mut, we'll get an `Incomparable` exception. *)
```
o-: I don't understand this. If you only ever call try_add on Decl_const declaration instructions this should never fail.

In fact you should probably refrain from using Scope.infer altogether since it does not seem necessary. You only care about the scope of const vars. And you keep track of it yourself (add on Decl_const and remove on drop). So you implicitly know the scope already without the inference.

transform_constantfold.ml
```diff
@@ -1,111 +1,130 @@
...
-             with our successors. *)
-          let succs = successors_at instrs pc in
-          find_targets instrs x (succs @ rest) (PcSet.add pc acc)
-      end
+    | [] -> Some instrs
+    | (_, pc) :: rest when PcSet.mem pc seen -> work instrs rest seen
+    | (penv, pc) :: rest ->
+      let scope = Scope.infer inp in
+      let (penv, instr) = normalize instrs.(pc) scope.(pc) penv in
+      let succs = List.map (fun pc -> (penv, pc)) (successors_at instrs pc) in
```
o-: There is some brittleness here. This works because a well-scoped program is guaranteed to have the same vars on all successors. It wouldn't work on an ill scoped program. The environments would not add up (of course we should assume well-scoped in the opts. but maybe a comment?)

More importantly it wouldn't work for muts.

```
1 mut x = 1
2 branch e 3 4
3 x <- 3
4 print x             # static env for x is 1||3
```

Update: actually it's not true. Currently it doesn't work for consts. See main comment.

o- (commented):
Hey, thanks for the patch. I think it goes in the right directing.

Looking at your PR I came up with the following program which breaks your optimization pass for obvious reasons:
```
 const x = 1
 branch (0==1) b1 b2
b1:
 goto end
b2:
 drop x
 const x = 3
end:
 print x
```

This is a nasty program and it's bad that we didn't see it coming. It's not just your optimization! For example
```
 const x = 1
 branch (0==1) b1 b2
b1:
 goto end
b2:
 drop x
 const x = 1
end:
```

Is broken by the cleanup pass as it removes the unused 2nd declaration which renders the program ill-scoped.

So I think we should probably fix it in the scope inference?

transform_constantfold.ml
mhyee: This is for the case `const x = v`, and I need to know if `v` is const or mut. It fails (with `Incomparable`) if `v` is mut.

I'll clarify in the comment.

transform_constantfold.ml
mhyee: I was hoping to avoid tracking the changed status (either threading the variable everywhere, or using a mutable reference). In fact, this proposal was designed to avoid that, but this was before the new API.

But I guess it's unavoidable now, so I'll fix it.

transform_constantfold.ml
mhyee: I could probably move this to `try_add`, so we only infer the scope if we are in the case `const x = v` and we need to determine if `v` is const or mut. (Right now we always infer the scope.)

transform_constantfold.ml
mhyee: I'll add a comment, since we assume well-scoped programs.

The example here would work, in that the optimization does nothing because `x` is mut. But the example in the main comment would definitely break.

transform_constantfold.ml
o-: oh, now I understand. But couldn't you just wait for the next iteration? Because constant propagation will replace v with the actual constant first, then you can check afterwards. No need to look up the variable!

transform_constantfold.ml
o-: yes the mut example was more to show how a naive extension of the reach  of the algo could break it.

transform_constantfold.ml
o-: This definitely feels odd. With the same argument you could lookup `v` in `print v` to see if its constant.

mhyee: Here's the case you're referring to, where we don't need to check the variable modes:
```
const x = 5    # Add x -> 5 to env
const y = x    # Rewrite: const y = 5
               # Add y -> 5 to env
print y        # Rewrite: print 5
```
But here's the case where we need to check the modes:
```
  call z = 'foo (1, 2)
function foo (const x, mut y)
  const a = x    # Can't rewrite, but can add a -> x to env
  const b = y    # Can't do anything because y is mut
  print a        # Rewrite: print x
  print b
  return 0
```

I don't understand your other comment. In the case `print v`, I don't care if `v` is const or mut. All I care is whether `v` has a mapping in the environment. (The existence of a mapping implies that `v` is constant. But `v` being constant does not imply a mapping exists, since `v` could be a const parameter.)

mhyee:
I pushed a new commit. (I'll rebase before merging.)

For the main comment about optimizations being horribly broken, I'll open a separate issue.

transform_constantfold.ml
o-: Aha, ok sorry for being stupid. So the case you can handle with this additional thing is, if you know something is const, but it's not a compile time constant then you can rename to the name of the runtime constant.

Could you maybe add that knowledge to the env instead? So the environment would be a mapping from names to `| RuntimeConst | CompiletimeConst of Lit`?

Sorry for pushing this, but I think longterm the more we can remove reliance on flow-sensitive analysis the easier it will be to prove optimizations. And this seems like an easy to resolve thing. But maybe again I just fail to see why it's necessary.

mhyee: There is a reason it failed your smell test! I can just check `formals`, which is a `ModedVarSet`. Now the code is a bit shorter and simpler. (And hopefully the comments still explain the situation.)

o-: Ah, you only consider function args. A more universal approach would also keep track of const aliases... E.g.:
```
function (const x)
  const y = x
  const z = y
  print z           # -> print x
```

This is what I actually had in mind:
https://gist.github.com/6f44ee64103fc0174dac8dcb285ce146

mhyee: Urgh, I totally missed that case. Thanks for the patch.

mhyee: Turns out that case is already handled by tracking only the parameters:
```
function (const x)
  const y = x   # add y -> x to env
  const z = y   # rewrite: const z = x
                # add z -> x to env
  print z       # rewrite: print x 
```
But your suggested patch is more explicit and general, and I already made the changes, so let's keep it.

mhyee:
I made another commit, just so the code exists in source control somewhere.

As we discussed today, I'll shelve this PR for now until we resolve #98.

mhyee:
Closing this and starting from scratch, with the new constant folding framework implemented in #119.

PR #100, opened 2017-05-02
[WIP] towards ssa consts

o-:


gasche:
If I understand correctly, in your proposal the index `{n}` in `x{n}` is erased at runtime, we map `x` in the environment. In particular, `x{n} = phi(x{i},x{j})` is valid but `x{n} = phi(x{i},y{j})` is not, as the latter cannot be interpreted by a no-op as you suggest (it would require renaming variables in the environment).

o-:
@gasche Yes that's the idea. It would allow us to treat phi as a nop.

The ssa is just an explicitly enforced property of the graph. We could even do without the indices and with a more clever scope inference, but I think explicit is nicer. Now the question is just if we can encode the properties of an SSA graph as a scope judgment.

doc/ir-semantics.text
```diff
@@ -217,71 +198,132 @@ extended with the new heap fragment into a new heap `H'`.
...
+      ∅
+  | xⁿ[e₁] <- e₂       ->
+      xⁿ :: vars(e₁) ∪ vars(e₂)
+  | xⁿ <- e            ->
+      xⁿ :: vars(e)
+  | print e
+  | return e
+  | clear x
+  | drop x             ->
+      [x*]  // TODO Better notation, requires some xⁿ in scope
```
gasche, line 258: The syntax uses `drop x{n}` instead of just `drop x`.

doc/ir-semantics.text
```diff
@@ -217,71 +198,132 @@ extended with the new heap fragment into a new heap `H'`.
...
+
+
+  Ai(v,l) = ⋃ merge { Ao(v,l') : l' ∈ pred(P(v),l) }
+  requires i ⊆ dom(Ai(v,l))
+  requires_typed i ⊆ Ai(v,l)
+  Ao(v,l) = Ai(v,l) ∪ declares i \ removes i
+  A, P ⊨d i : Ai(v,l)
   pred(P(v), l) = ∅  ⇒  i = start
   -----------------------------------------------------------
   A, P ⊨d v, l → i
```
gasche, line 313: You would have to update that conclusion with A{i}, A{o}.

gasche:
I'm not sure about the new scope judgment. One may to model phi-nodes could be to say that, in general, scopes A are not just set of variables in scope, but sets of *consistent sets* of variables, where a set is consistent if it only contain versions of the same variable. So if you have something like

```
  const x^1 = 1
  goto merge
  const x^2 = 2
  goto merge
  x^3 = phi(x^1, x^2
```

I would expect the following scope assignments:

```
  [] const x^1 = 1 [x:{x^1}]
  [x:{x^1}] goto merge [x:{x^1}]

  [] const x^2 = 2 [x:{x^2}]
  [x:{x^2}] goto merge [x:{x^2}]

  [] merge: [x:{x^1, x^2}]
  [x:{x^1, x^2}] x^3 = phi(x^1, x^2) [x:{x^3}]
```

then, the instruction judgment would impose that, for all instructions that are neither labels no phi-nodes, the input and output scopes only have singletons in their scope.

o-:
> One may to model phi-nodes could be to say that, in general, scopes A are not just set of variables in scope, but sets of consistent sets of variables, where a set is consistent if it only contain versions of the same variable

I like that formulation

> then, the instruction judgment would impose that, for all instructions that are neither labels no phi-nodes, the input and output scopes only have singletons in their scope.

Yes I think my version is broken as it would allow:
```
 x1       |        x2
       print x1
      x3 = phi(x1, x2)
```

gasche (commented):
(I realize now that the review comments below were pending, so I'm sending the review because why not.)

gasche:
The consensus is to go with the simplification implemented in #121 instead of SSA, so closing this.

PR #102, opened 2017-05-02
Document `drop` and `clear` in the specification

gasche:


o-:
thanks, lgtm

o-:
closes #102

doc/ir-semantics.text
```diff
@@ -177,6 +179,14 @@ extended with the new heap fragment into a new heap `H'`.
        = (I, T, (H,E)[x ← v], E, succ I l)
     when v := eval P H E e
 
+  step   (clear x) (P, I, T, H, E, l)
+       = (I, T, (H,E)[x ← ⊥], E, succ I l)
+    when v := eval P H E e
```
mhyee, line 184: This line should be removed.

gasche: Good catch, fixed in 2c0fb2e.

PR #103, opened 2017-05-02
Array

aviralg:
Added arrays. #85

### DIFF
--- a/eval.ml
+++ b/eval.ml
@@ -307,6 +307,16 @@ let reduce conf =
        heap = update conf.heap conf.env x v;
        pc = pc';
      }
+  | Array_assign (x, i, e) ->
+    let vi = eval conf i in
+    let ve = eval conf e in
+    let arr = lookup conf.heap conf.env x in
+    let vs = get_array arr in
+    vs.(get_int vi) <- ve;
+    { conf with
+      heap = update conf.heap conf.env x (Array vs);
+      pc = pc';
+    }
   | Branch (e, l1, l2) ->
      let b = get_bool (eval conf e) in
      { conf with pc = resolve conf.instrs (if b then l1 else l2) }
--- a/eval.ml
+++ b/eval.ml
@@ -27,12 +27,13 @@ type configuration = {
   continuation : continuation list;
 }
 
-type type_tag = Nil | Bool | Int | Fun_ref
+type type_tag = Nil | Bool | Int | Fun_ref | Array
 let get_tag : value -> type_tag = function
   | Nil -> Nil
   | Bool _ -> Bool
   | Int _ -> Int
   | Fun_ref _ -> Fun_ref
+  | Array _ -> Array
 
 exception Unbound_variable of variable
 exception Undefined_variable of variable
@@ -85,7 +86,7 @@ let clear heap env x =
   | Val _ -> raise Invalid_clear
   | Ref a -> Heap.add a Undefined heap
 
-let value_eq (v1 : value) (v2 : value) =
+let rec value_eq (v1 : value) (v2 : value) =
   match v1, v2 with
   | Nil, Nil -> true
   | Nil, _ | _, Nil -> false
@@ -94,11 +95,19 @@ let value_eq (v1 : value) (v2 : value) =
   | Int n1, Int n2 -> n1 = n2
   | Int _, _ | _, Int _ -> false
   | Fun_ref f1, Fun_ref f2 -> f1 = f2
+  | Fun_ref _, _ | _, Fun_ref _ -> false
+  | Array vs1, Array vs2 ->
+    let forall2 p arr1 arr2 =
+      Array.mapi (fun i v1 -> p v1 arr2.(i)) arr1
+      |> Array.for_all (fun b -> b)
+    in
+    Array.length vs1 = Array.length vs2
+    && forall2 value_eq vs1 vs2
 
 let value_plus (v1 : value) (v2 : value) =
   match v1, v2 with
   | Int n1, Int n2 -> n1 + n2
-  | (Int _ | Nil | Bool _ | Fun_ref _) as x1, x2 ->
+  | (Int _ | Nil | Bool _ | Fun_ref _ | Array _) as x1, x2 ->
       let expected = (Int, Int) in
       let received = get_tag x1, get_tag x2 in
       raise (ProductType_error { expected; received })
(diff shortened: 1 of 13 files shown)


gasche:
@o- in the current version of this PR, we implemented array assignment in the obvious way, namely by translating to OCaml array assignment:

```ocaml
  | Array_assign (x, i, e) ->
    let vi = eval conf i in
    let ve = eval conf e in
    let arr = lookup conf.heap conf.env x in
    let vs = get_array arr in
    vs.(get_int vi) <- ve;
    { conf with
      heap = update conf.heap conf.env x (Array vs);
      pc = pc';
    }
```

This does not respect the current by-value semantics, because the unique-owner assumption we make when we update-in-place instead of copying the array here is broken by function calls.

We could fix it by doing a copy here, or by doing a copy of values when evaluating a function call (but it's weird to have code there), or by changing the semantics to have the value of an array be a reference to its content.

What do you think?

eval.ml
```diff
@@ -94,11 +95,19 @@ let value_eq (v1 : value) (v2 : value) =
   | Int n1, Int n2 -> n1 = n2
   | Int _, _ | _, Int _ -> false
   | Fun_ref f1, Fun_ref f2 -> f1 = f2
+  | Fun_ref _, _ | _, Fun_ref _ -> false
+  | Array vs1, Array vs2 ->
```
mhyee, line 99: Minor: could you add a case for `Array _, _ | _, Array _`? I know it's redundant (but some of the other patterns are already redundant), but it makes the case explicit and then you won't forget to add it if you add another value type.

instr.ml
```diff
@@ -167,6 +176,7 @@ let value_of_string : string -> value = function
   | n ->
     try Int (int_of_string n) with _ ->
       Printf.kprintf invalid_arg "value_of_string %S" n
+(* TODO add case for array *)
```
mhyee, line 179: Will this be added in a later PR?

parser.mly
```diff
@@ -7,6 +7,7 @@
 %token LPAREN RPAREN LBRACKET RBRACKET LBRACE RBRACE
 %token COLON EQUAL LEFTARROW TRIPLE_DOT COMMA
 %token CONST MUT BRANCH GOTO PRINT OSR STOP READ DROP CLEAR RETURN CALL VERSION FUNCTION
+%token ARRAY LENGTH
```
mhyee, line 10: Don't you need to update the parser error messages?

transform_constantfold.ml
```diff
@@ -43,6 +43,9 @@ let const_prop ({formals; instrs} : analysis_input) : instructions option =
     | Assign (y, e) ->
       assert (x <> y);
       Assign (y, replace e)
+    | Array_assign (y, i, e) ->
+      assert (x <> y);
```
mhyee, line 47: I'm a bit suspicious. (I guess this is partly because constant folding is being changed in a new PR, and we're also considering changing the semantics of arrays.)

I believe the assert will be triggered in the following program:
```
const x = [1]
x[0] <- 0
```

The function `find_candidates` will see the constant definition `const x = [1]` and add it as a "candidate," and then try to do the constant propagation here, where the assert will trigger.

mhyee (changes requested):
I'm not convinced arrays and constant folding will play nicely. I made a comment about this and gave an example. You could also try running `--opt=all` on the array examples and see what happens.

Other than that, the rest of the implementation looks fine.

mhyee:
See also the discussion [here](https://github.com/reactorlabs/sourir/pull/85#issuecomment-297060236).

I would prefer changing the semantics. I think this will also address the concerns I have with constant propagation.

Also, silly/stupid example, but is the following currently legal?

```
const len = length([1, 2, 3])
const val = [1, 2, 3][1]
[1, 2, 3][0] <- 0
```

instr.ml
gasche: Yes, I'll play with the parser plugging to call the parser directly here.

transform_constantfold.ml
gasche: With the current array semantics (see @o-'s comment https://github.com/reactorlabs/sourir/pull/85#issuecomment-297063689 ) it is illegal to update a `const`-declared array, so we could reject this program at compile-time anyway (we did not think of adding this to `check`, though, and it would become obsolete if we go for a by-reference semantics).

gasche:
> Also, silly/stupid example, but is the following currently legal?
> 
> ```
> const len = length([1, 2, 3])
> const val = [1, 2, 3][1]
> [1, 2, 3][0] <- 0
> ```

This is rejected by the ANF-like structure of expressions. Array literals are expressions but not "simple expressions", so they cannot be used as parameters of `length`, assignment or indexing. Furthermore, the grammar for the assignment instruction currently require that the array be passed as a variable (instead of a simple expression as one might expect), following @o-'s choice in the specification, but this doesn't make a difference as variables are the only simple expressions of array type (literals and function names cannot be arrays).

eval.ml
gasche: I don't know of other patterns that are redundant -- in general redundant patterns are programming errors so it's good to heed the warnings and remove them. This idiom is the only one I know where it is natural to write a redundant pattern. If we want to have the last case, I would rather have it commented out to not get the warning. Otherwise I think that expecting the next person that adds a new case to recognize the pattern and do the right thing is not unreasonable.

o-:
sorry I didn't get around looking at it. too many changes are in flight right now and i am not sure which decisions depend on which. I believe that we should first agree on what kind of variables we have and how we use them. before that it just doesn't make sense do discuss what a 'const array' is...

anyways in principle I agree with the in place update of arrays.

eval.ml
mhyee: I'm fine with leaving it as is -- maybe I was just reading it too quickly and thought some patterns wer redundant. And if someone forgets a case, the compiler should complain.

parser.mly
gasche: I'll do the update, but only after we converge on a syntax and merge something, so that the design has stabilized.

gasche:
I merged this based on the discussion in #104. We must be vigilant to the fact that the semantics does not exactly match the current specification right now, and some things like interaction with optimizations may need to be changed/polished. But I think that this codebase is a solid prefix of further changes that we can build on, and that having it in the master will help us iterate faster on those.

I'll implement the missing `value_of_string` case and update the parser messages now.

PR #106, opened 2017-05-03
Cleanup

gasche:
Minor things coming from my work with @aviralg on #103. @o-, you may or may not be interested in this mode of use of `%a` in printers (which require printing functions to follow a certain type/protocol, taking a buffer first and then the object to print).


mhyee (approved):

o-:
nice

Issue #104, opened 2017-05-03
Simplification proposal


gasche:
Here would be a proposal for a great simplification:

- move `const x` to mutable `var x` as suggested, only bound in the
  lexical environment (no heap allocation, no aliasing, no `&x`).

- remove `mut x` and let people use 1-arrays instead when they want an
  aliasing semantics. Only use the heap to allocate arrays
  (as previously proposed, an array *value* is `array a` where `a` is
  a heap adress to a sequence of values).

- require that each variable has a single definition in each segment

I can live with that.

o-:
> remove mut x and let people use 1-arrays instead

😺

o-:
I like how you formalized the proposal I wanted to and I formalized the one you wanted to.

This is very close to what I planned to implement (minus the great idea of removing mut). Should I run with the implementation or are you working on it?

gasche:
Feel free to go ahead and implement. But maybe we could try to merge arrays first, given that they impact the design?

gasche:
I realize now that in the other PR you suggested to sort out variables first and only merge arrays later. I'm fine with either. The current array PR is easy to change to get a by-reference semantics as I previously proposed, which is the one we need to replace `mut`. Doing this after a "great simplification" PR would be a bit more work because of conflicts, but I'm fine with dealing with the conflicts.

o-:
On 05/03/2017 07:23 PM, Gabriel Scherer wrote:
> I realize now that in the other PR you suggested to sort out variables first and only merge arrays later. I'm fine with either. The current array PR is easy to change to get a by-reference semantics as I previously proposed, which is the one we need to replace `mut`. Doing this after a "great simplification" PR would be a bit more work because of conflicts, but I'm fine with dealing with the conflicts.

it's ok. Go ahead and merge the arrays if you all are happy. It's one
thing less dangling! Probably less headache this way round.

mhyee:
Yeah, just merge something first because there's too many proposals in flight right now.

Anyway, I like this proposal. It seemed a bit weird to me at first, but it makes sense to keep non-array variables on the stack/registers/not-heap, and arrays on the heap.

Minor comment about aesthetics: can we declare arrays with `arr a` so it's only three letters, to match with `var x`?

mhyee:
Fixed by #121.

mhyee:
Re-opening this. We implemented the simplification, but we haven't done the third part yet:

> require that each variable has a single definition in each segment

gasche:
[...]

PR #107, opened 2017-05-04
[WIP] grand simplification

o-:
That's as far as I got with #104 

I abandoned the heap and decided to store arrays on the ocaml heap. It's maybe conceptually not as nice but it saves a lot of loc


gasche:
I am worried about dropping the heap, because I think that reasoning about it may matter in correctness proofs (it remains part of the machine state), and it's nicer if we can reason directly on the state as manipulated by the implementation. On the other hand, if we decide right now to not have any optimization of heap-allocation scenarios, then that's fine.

o-:
tried to reply by mail. that didn't work

> I am worried about dropping the heap,

I agree. I'll try something else. The main problem is that allocating arrays in an expression makes eval have some side-effects (ie. growing the heap). I am not sure how to deal with it. We could
* make creating an array an instruction. This keeps eval pure, but is a bit ugly.
* have array as values as the result of the expression and then the assignment instruction creates the heap cell. Eval stays pure, but it's a bit tricky to get right and not have the array values escape to some unforeseen places.
* Argue that extending the heap is not an observable side-effect (since the fresh reference cannot have escaped) and allow it.

mhyee:
The third option seems slightly better than the first. The second seems brittle.

gasche:
I would think that the first plan is fine: `alloc x[n]` assigns a fresh array of `nil`s of size `n` to `x`.

mhyee:
How would you write array literals? `alloc x[] = [1, 2, 3]`?

What if you wanted to use an array literal within an expression? Or do we want to forbid this?

Actually, if we use the third option, doesn't the array reference escape?
```
  call x = foo([1, 2, 3])
function foo(ref a)
  return a
````

So maybe we want to use the first option, and forbid arrays in expressions. This would be similar to how we handle calls: `call x = foo (...)`.

gasche:
Your `alloc x[] = [...]` proposal seems fine to me.

Array literals todays are expressions but not simple expressions -- they cannot appear as argument of a function, for example, so the further issues you mention are already ruled out.

aviralg:
If we allow side effects in expressions, what do we lose? Does it make it difficult to prove correctness?

gasche:
I think that it does not introduce an essential difficulty, but it does make everything a bit heavier, in the implementation and in the proof, as expression-related functions (and reasonings) need to return a new updated heap.

o-:
> but it does make everything a bit heavier,

yeah it was really painful in #110 and I think not worth pursuing after all...

I still find option 2 (have array as values as the result of the expression and then the assignment instruction creates the heap cell) the most attractive. But I don't know how to implement it :(

o-:
So I am keeping this PR open to keep the discussion visible. But I think it's better to start from scratch. This PR is not good because it uses the ocaml heap. #110 is painfully dragging the heap around in eval and is doing too much anyways. Maybe it has to be done in steps. For example first remove const. Then reintroduce local vars. I don't know

gasche:
Done in #121.

PR #110, opened 2017-05-05
[WIP] grand simplification attempt 2

o-:
This is an even more opinionated simplification attempt implementing (among other things) #104 

TODO
* parsing of arrays (needs access to the heap)
* fix all optimizations
* fix rest of examples
* fix tests
* a static differentiation between a heap extension and a heap modification would be nice!


aviralg:
I am working on this now, right from where you left off.

aviralg:
Stopped working on it. Didn't get too far. Will work on inlining instead.

PR #111, opened 2017-05-05
Run examples in a shell script

### LINKED ISSUE #108: Travis Doesn't fail
Currently `make test_examples` breaks but travis is still green. Travis should fail in this case!

mhyee:
Resolves #108.

Each example is tested separately, and the result is saved. If any of the tests fail, then the script returns a non-zero exit code. The script will also print the name of the failing example.

So Travis should see the failure (and the build should fail with this PR).

I also stuck `test_examples` in `make all`. This was probably why we didn't catch this error earlier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,7 @@ OCAMLBUILD=ocamlbuild -use-ocamlfind -use-menhir -no-links
 # -no-links: do not create symlink from build outputs in _build into
 #            the project directory
 
-all: tests sourir
-	rm -rf $(TEMPDIR)
+all: tests test_examples sourir
 
 tests:
 	$(OCAMLBUILD) tests.byte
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -27,16 +27,7 @@ run: sourir
 TEMPDIR := $(shell mktemp -d)
```
gasche: can we get rid of this then?

gasche (commented):
Looks nice; it seems that it should also simplify the side-effectfulness of temporary directory creation.

Makefile
mhyee: Good catch, I'll remove it.

mhyee:
The side-effectfulness of the tempdir occurs in the bash script, but at least it's out of the makefile.

I confirmed that Travis properly runs the script and fails (because `array.sou` is currently broken). I also confirmed (in a deleted test commit) that "fixing" `array.sou` allows Travis to pass.

So I'd say this is ready to merge, whenever you want to click the button.

gasche:
We need to understand how to fix `array.sou` (it may be a bug in the implementation). I'll fix it and then merge here, thanks!

gasche:
I fixed the example in #112 -- it was a bug in the `make_const` pass. I just restarted the Travis tests here.

o-:
thanks for fixing my bugs.

PR #116, opened 2017-05-10
Add operators for sub, mult, div, mod

mhyee:
This PR adds the following operators: `- * / % < <= > >= && ||`. It also adds the unary operators `- !`.

Note that primops are no longer pure, because `/` and `%` can raise an `Eval.Division_by_zero` exception. I made this an explicit sourir exception, rather than falling back to OCaml's exception.


mhyee:
This is now ready for review.

gasche:
Ok to merge once conflicts are fixed.

We should factorize the code in the `value_op*` operations. I can do that after merging.

You are a bit too fond of `when`-guards, I think. They are a tempting feature as they read nicely, but in my experience they can often make the code more fragile than it should (they make it harder to reason about exhaustiveness of pattern-matching, see 7063e8ae13a61cb45fb23343495b202b29da3370 for example). When it is easy to use a single clause with an if/then/else inside, it is the better move.

mhyee:
Rebased. I also removed the `when`-guards in favor if `if-then-else`. Travis is running the tests right now.

FYI, there appears to be a leftover warning from your arrays work:
```
File "transform_constantfold.ml", line 125, characters 4-486:
Warning 8: this pattern-matching is not exhaustive.
Here is an example of a case that is not matched:
Decl_array (_, _)
```

gasche:
Yeah, I saw that as well, and fixed it in ongoing work. I'll sent more PRs as the simplification work progresses.

PR #119, opened 2017-05-12
Better constant folding

gasche:
This patch extends constant-folding to work correctly over both `const` and `mut` binding: if an assignment is done on a mutable variable, its previous static value (if it exists) is removed from the dictionary of known approximations.

This is the patch that I had in mind when I proposed to "not change anything to the language" two Mondays ago.  By using a flow analysis in the constant folding, we can nicely handle situations a merge point has to merge environments with distinct approximations for a given variable.

I am still sentimental about moving from this nice place, but the reason I did the patch is that it's on the least-effort path towards the big simplification anyway: I started hacking on the simplification, but without this patch constant propagation completely breaks down when adding assignments to variable, and fixing the tests becomes impractical.

### DIFF
--- a/transform_constantfold.ml
+++ b/transform_constantfold.ml
@@ -12,18 +12,8 @@ open Instr
  * used, and the declaration can be removed by running `minimize_lifetimes`.
  *)
 let const_prop ({formals; instrs} : analysis_input) : instructions option =
-  (* Finds the declarations that can be used for constant propagation.
-     Returns a list of (pc, x, l) where `const x = l` is defined at pc `pc`. *)
-  let rec find_candidates instrs pc acc =
-    if pc = Array.length instrs then acc
-    else match[@warning "-4"] instrs.(pc) with
-    | Decl_const (x, Simple(Constant(l))) ->
-        find_candidates instrs (pc+1) ((pc, x, l) :: acc)
-    | _ -> find_candidates instrs (pc+1) acc
-  in
-
   (* Replaces the variable `x` with literal `l` in instruction `instr`. *)
-  let convert x l instr =
+  let replace x l instr =
     let replace = Edit.replace_var_in_exp x l in
     let replace_arg = Edit.replace_var_in_arg x l in
     match instr with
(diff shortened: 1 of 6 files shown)


transform_constantfold.ml
```diff
@@ -69,53 +62,91 @@ let const_prop ({formals; instrs} : analysis_input) : instructions option =
...
+        VarMap.add x (Value l) cur
+      | Decl_array (x, _) ->
+        (* this case could be improved with approximation for arrays so
+           that, eg., length(x) could be constant-folded *)
+        VarMap.add x Unknown cur
+      | Decl_const (x, _) | Decl_mut (x, _) ->
+        VarMap.add x Unknown cur
+      | Call (x, _, _) as call ->
+        cur
+        |> ModedVarSet.fold (fun (_, x) -> VarMap.remove x) (changed_vars call)
```
mhyee: Is there a reason the changed variables are removed, rather than setting them to `Unknown`? In other words, what is the difference between `Unknown` and not being in the map? (More comments would be helpful for answering the second question.)

transform_constantfold.ml
```diff
@@ -69,53 +62,91 @@ let const_prop ({formals; instrs} : analysis_input) : instructions option =
...
+        | Some (Value v1), Some (Value v2) ->
+          if Eval.value_eq v1 v2 then Some (Value v1) else Some Unknown
+      in
+      if VarMap.equal Approx.equal cur incom then None
+      else Some (VarMap.merge merge_approx cur incom)
     in
-    Some (List.fold_left propagate instrs candidates)
+    let update pc cur =
+      match[@warning "-4"] instrs.(pc) with
+      | Decl_const (x, Simple(Constant l))
```
mhyee: Nitpick: missing space in `Simple (Constant l)`

mhyee (commented):
To clarify, this is actually constant *propagation*, since the code doesn't handle cases like `const x = (1 + 2)`. It also doesn't handle cases like `const x = y`.

That's OK for now, I can add the other cases later.

Could you also add a quick test case with muts, since the existing tests only consider consts? Something like:

```
  mut x = 1
  branch (1==1) l1 l2
l1:
  print x    # print 1
  x <- 2
  goto fin
l2:
  x <- 3
  print x  # print 3
fin:
  stop x
```

transform_constantfold.ml
gasche: This was a (minor) bug, it did not break anything except the invariant that the domain of the map is exactly the set of variables in scope. I fixed it. Thanks for the catch!

gasche:
I added two extra examples as recommended, thanks!

mhyee (approved):
Looks great, thanks! You can merge when Travis is green.

PR #114, opened 2017-05-09
Add some interesting examples from tests.ml

mhyee:
WIP,  because some of these examples are broken. From a quick glance, they appear to be bugs in sourir and not the examples.

`Fatal error: exception Assert_failure("transform_hoist.ml", 182, 6)`:
```
Example cf_01_after.sou failed!
Example cf_01_before.sou failed!
Example min_live_01_after.sou failed!
Example movedrop_01_before.sou failed!
Example movedrop_02_after.sou failed!
Example movedrop_02_before.sou failed!
Example movedrop_03_after.sou failed!
Example movedrop_03_before.sou failed!
Example movedrop_04_before.sou failed!
```

`Fatal error: exception Invalid_argument("index out of bounds")`:
```
Example cf_03_after.sou failed!
Example movedrop_04_after.sou failed!
```

`Fatal error: exception Eval.Unbound_variable("i")`:
```
Example double_loop.sou failed!
```

PR #120, opened 2017-05-12
fix a hoist_drop bug due to a fallthrough

gasche:
`Transform.remove_fallthroughs_to_label` has been removed by
f86504b38b6e1ff4d93a1bf0dfe194ef48f7a2e0, but I'm not sure why, it is
still necessary for examples/regression_without_fallthrough.sou to work.

Besides, in the case where there is a fallthrough, the assertion in `is_blocking`:

    let blocking = VarSet.mem var (required_vars instr) in
    let declared = VarSet.mem var (ModedVarSet.untyped (declared_vars instr)) in
    assert (blocking || not(declared));
    blocking

is violated: in the program

    var x = 10
    label:
    drop x

the instruction above the label (`var x = 10`) is not blocking
(it annihilates instead) but it is declaring `x`. I removed the
assertion -- I don't understand what it protects again, and it
caused an inconvenience while debugging this.

### DIFF
--- a/examples/regression_without_fallthrough.sou
+++ b/examples/regression_without_fallthrough.sou
@@ -1,0 +1,4 @@
+const x = 10
+label:
+drop x
+print 1
--- a/transform_hoist.ml
+++ b/transform_hoist.ml
@@ -1,5 +1,21 @@
 open Instr
 
+let remove_fallthroughs_to_label instrs =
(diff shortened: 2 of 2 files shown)


mhyee:
Looks like this will fix the firing asserts in #114.

mhyee (approved):

PR #121, opened 2017-05-12
The great simplification

gasche:
Builds on #120 which fixes a bug revealed by the change.


gasche:
We still have to update the operational semantics description, but I'm calling it a day.

doc/ir-semantics.text
```diff
@@ -18,17 +18,17 @@ Reserved Names:
...
-  | const x
+  | var x
   | mut x
 
   a  addresses
 
   i ::=    instructions
-  | const x = e                   constant variable
-  | mut x = e                     mutable variable
+  | var x = e                     constant variable
```
mhyee: Delete "constant"

doc/ir-semantics.text
```diff
@@ -125,74 +105,50 @@ Update (partial) function, returns an H:
   eval P H E se          = simple-eval P H E se
   eval P H E primop(se*) = 〚primop〛(v*)
     where (v := simple-eval P H E se)*
-  eval P H E [se*]       = [v*]
-    where (v := simple-eval P H E se)*
   eval P H E x[se]       = vₙ
```
mhyee: Subscript should be `i`

doc/ir-semantics.text
```diff
@@ -125,74 +105,50 @@ Update (partial) function, returns an H:
...
-
-`replace-env H E def*` creates a new lexical environment `E'` from
-scratch (it does not extend the input environment `E`) from the list
-of definitions `def*`, interpreted/evaluated in `E`. The heap `H` is
-extended with the new heap fragment into a new heap `H'`.
-
-  replace-env H E def* := (H, ∅) + (build-env H E def)*
+    where a := E[x]
+      and i := simple-eval P H E se
+      and H[a] = [v₁..., vᵢ, ...]
```
mhyee: Nitpick: comma after `v₁`

doc/ir-semantics.text
```diff
@@ -125,74 +105,50 @@ Update (partial) function, returns an H:
...
-       = (I, T, H', E', succ I l)
-   where (H', E') := extend-env H E (const x = e)
-
-  step   (mut x = e) (P, I, T, H, E, l)
-       = (I, T, H', E', succ I l)
-   where (H', E') := extend-env H E (mut x = e)
+  step   (var x = e) (P, I, T, H, E, l)
+       = (I, T, H, E', succ I l)
+   where v := eval P H E e
+     and E' := E[x ↦ eval P H E e]
```
mhyee: `E[x ↦ v]` since you already `eval` on the line above.

doc/ir-semantics.text
```diff
@@ -18,17 +18,17 @@ Reserved Names:
...
   a  addresses
 
   i ::=    instructions
-  | const x = e                   constant variable
-  | mut x = e                     mutable variable
+  | var x = e                     constant variable
   | drop x                        drop a variable from scope
-  | clear x                       de-initialization of a mutable
   | x <- e                        assignment
+  | array x[n]                    array allocation
```
mhyee: Don't we allow using an expression for the array length? So `array x[e]`?

doc/ir-semantics.text
```diff
@@ -125,74 +105,50 @@ Update (partial) function, returns an H:
...
-
-  step   (clear x) (P, I, T, H, E, l)
-       = (I, T, (H,E)[x ← ⊥], E, succ I l)
+     and E' := E[x ↦ v]
 
   step   (drop x) (P, I, T, H, E, l)
        = (I, T, H, E', succ I l)
     when E' := E \ {x}
 
+  step    (array x[n]) (P, I, T, H, E, l)
```
mhyee: Again, array length is specified by an expression.

doc/ir-semantics.text
```diff
@@ -125,74 +105,50 @@ Update (partial) function, returns an H:
...
+        = (I, T, H', E', succ I l)
+   when a fresh
+     H' := H[a ↦ [nil⁽ⁿ⁾]]
+     E' := E[x ↦ a]
+
+  step    (array x = [e₁, ..., eₙ]) (P, I, T, H, E, l)
+        = (I, T, H', E', succ I l)
+   when a fresh
+     (vᵢ := eval P H E eᵢ)ⁱ
```
mhyee: Superscript should be `n`.

doc/ir-semantics.text
mhyee: Also, for the array index, do we allow an expression or does it have to be a simple expression?

doc/ir-semantics.text
```diff
@@ -125,74 +105,50 @@ Update (partial) function, returns an H:
...
-    when n  := eval P H E eᵢ
-     and vₐ := eval P H E eₐ
-     and v  := (H,E)[x]
-     and v   = [v₁..., vₙ, ...]
-     and v' := [v₁..., vₐ, ...]
+       = (I, T, H', E, succ I l)
+    when i := eval P H E eᵢ
+     and v := eval P H E eₐ
+     and a  := E[x]
+     and H[a] = [v₁..., vᵢ, ...]
```
mhyee: Nitpick: Comma after `v₁`. Same for the line after.

doc/ir-semantics.text
```diff
@@ -241,19 +197,14 @@ extended with the new heap fragment into a new heap `H'`.
...
-requires i =
-  | mut x = e
-  | const x = e
-  | osr (e, f, lᵥ, l, (x = ee)*)
-                    -> vars(e) ∪ vars(ee)*
-  | x <- e          -> x :: vars(e)
-  | branch x _
-  | print x         -> [x]
-  | _               -> []
+requires i = bound_vars(i)
```
mhyee, line 208: Was `bound_vars` defined anywhere?

check.ml
```diff
@@ -99,19 +79,15 @@ let well_formed prog =
         | Simple e -> check_simple_expr e
         | Op (_op, xs) ->
           List.iter check_simple_expr xs in
-      let check_arg = function
-        | Arg_by_val e -> check_expr e
-        | Arg_by_ref x -> () in
+      let check_arg = check_expr in
       let check_osr = function
```
mhyee, line 83: Could you write this as `let  check_osr (Osr_var (_, e)) = check_expr e in`?

check.ml
```diff
@@ -99,19 +79,15 @@ let well_formed prog =
         | Simple e -> check_simple_expr e
         | Op (_op, xs) ->
           List.iter check_simple_expr xs in
-      let check_arg = function
-        | Arg_by_val e -> check_expr e
-        | Arg_by_ref x -> () in
+      let check_arg = check_expr in
```
mhyee, line 82: Why not just use `check_expr` instead of defining `check_arg`?

disasm.ml
```diff
@@ -42,11 +42,7 @@ let disassemble_instrs buf ?(format_pc = no_line_number) (prog : instructions) =
       | Op (Array_length, [array]) -> pr buf "length(%a)" simple array
       | Op ((Array_index | Array_length), _) -> assert(false)
     in
-    let dump_arg buf arg =
-      match arg with
-      | Arg_by_val e      -> dump_expr buf e
-      | Arg_by_ref x      -> pr buf "&%s" x
-    in
+    let dump_arg buf arg = dump_expr buf arg in
```
mhyee, line 45: Why not just use `dump_expr` instead of defining `dump_arg`?

disasm.ml
```diff
@@ -72,10 +65,7 @@ let disassemble_instrs buf ?(format_pc = no_line_number) (prog : instructions) =
     | Read var                        -> pr buf " read %s" var
     | Osr {cond; target = {func; version; label}; map} ->
       let dump_var buf = function
-        | Osr_const (x, e)     -> pr buf "const %s = %a" x dump_expr e
-        | Osr_mut (x, e)       -> pr buf "mut %s = %a" x dump_expr e
-        | Osr_mut_ref (x, y)   -> pr buf "mut %s = &%s" x y
-        | Osr_mut_undef x      -> pr buf "mut %s" x
+        | Osr_var (x, e)     -> pr buf "var %s = %a" x dump_expr e
```
mhyee, line 68: Again, just wondering if the pattern matching can be done in the parameters. But I guess it might be cleaner this way.

edit.ml
```diff
@@ -118,18 +118,11 @@ let replace_uses_in_instruction old_name new_name instr : instruction =
     | Simple se -> Simple (in_simple_expression se)
     | Op (op, exps) ->
       Op (op, List.map in_simple_expression exps) in
-  let in_arg exp : argument =
-    match exp with
-    | Arg_by_val e -> Arg_by_val (in_expression e)
-    | Arg_by_ref x -> if x = old_name then Arg_by_ref new_name else Arg_by_ref x in
+  let in_arg e : argument = in_expression e in
   let in_osr osr : osr_def =
```
mhyee, line 122: Another place where pattern matching could be done in the parameter?

eval.ml
```diff
@@ -465,6 +409,11 @@ let read_trace conf = List.rev conf.trace
 let rec reduce_interactive conf =
   if stop conf then conf
   else begin
+    if false then begin
```
mhyee: Leftover from debugging?

instr.ml
```diff
@@ -194,43 +153,36 @@ let expr_vars = function
     List.map simple_expr_vars xs
     |> List.fold_left VarSet.union VarSet.empty
 
-let arg_vars = function
-  | Arg_by_val e -> expr_vars e
-  | Arg_by_ref x -> VarSet.singleton x
+let arg_vars = expr_vars
```
mhyee, line 156: Again, I feel like `arg_vars` is used in so few places that we might as well change it to `exp_vars`.

tests.ml
```diff
@@ -767,13 +737,13 @@ let do_test_const_prop_driver () =
     drop y
     stop 0
   |expect};
-  (* Test with booleans, updating mut vars, and branching *)
+  (* Test with booleans, updating var vars, and branching *)
```
mhyee: "updating ~~var~~ vars"

tests.ml
```diff
@@ -1666,12 +1485,20 @@ let suite =
    "pull_drop">:: do_test_pull_drop;
    "move_drop">:: do_test_drop_driver;
    "test_functions">:: test_functions;
-   "test_mut_to_const">:: do_test_mut_to_const;
    "deopt">:: do_test_deopt;
    "array">:: do_test_array;
    ]
 ;;
 
+(*
```
mhyee: Debugging leftover?

mhyee (changes requested):
I'd definitely call it a "great simplification" if we can lose ~600 LOC!

Could you add two more tests/examples? I want to confirm that we have the correct behaviour and writing to a param, and also demonstrate using arrays as out-params. Something like the following:

```
  var x = 42
  call y = foo (x)
  print x          # prints 42
  print y          # prints 0
  stop y
function foo (var x)
  x <- 0
  print x          # prints 0
  return x
```
```
  array x = [42]
  call y = foo (x)
  print x[0]          # prints 0
  print y[0]          # prints 0
function foo (var x)
  x[0] <- 0
  print x[0]          # prints 0
  return x
```

Also, a minor comment. In the semantics doc, is there a difference between using `when` or `where`?

doc/ir-semantics.text
gasche: Currently it has to be a simple expression, which is consistent with the fact that primitive arguments must be simple expressions.

doc/ir-semantics.text
gasche: No, I assume this is known.

check.ml
gasche: In the type I kept `argument` a separate type, currently a synonym of `expression`. I kept these aliases as well, in case we separate the two types again.

check.ml
gasche: Currently that would be correct. I am thinking of implementing `array x[e]` and `array x[] = [...]` in osr-map, so that we would have several cases again.

tests.ml
gasche: Indeed. Instead of removing them, I'll turn them into one-array tests.

gasche:
Thanks for the review (nice catches!), I rebased the PR with fixes.

doc/ir-semantics.text
```diff
@@ -125,74 +105,51 @@ Update (partial) function, returns an H:
...
-     and vₐ := eval P H E eₐ
-     and v  := (H,E)[x]
-     and v   = [v₁..., vₙ, ...]
-     and v' := [v₁..., vₐ, ...]
+       = (I, T, H', E, succ I l)
+    when i := eval P H E eᵢ
+     and v := eval P H E eₐ
+     and a  := E[x]
+     and H[a] = [v₁, ..., vᵢ, ...]
+     and H' := H[a ↦ [v₁..., v, ...]]
```
mhyee: Nitpick: missing comma and space after `v₁`

tests.ml
mhyee: Looks like this is still in the code and commented out?

mhyee (approved):
Just two minor comments, other than that, looks great!

PR #124, opened 2017-05-16
Examples from testsml

o-:
@mhyee I cannibalized your #114 and added the necessary bugfixes


mhyee (approved):
Thanks for the fixes!

gasche:
Merged! With these changes, coverage increases from 84.31% to 84.82%.

PR #125, opened 2017-05-16
Constant folding

mhyee:
I've lost count of what attempt number this is. This extends the constant propagation work in #119 with constant folding. I did not implement copy propagation.

A variable declaration introduces a constant `x` if it is of the form:
1. `var x = l` where `l` is a literal
2. `var x = op(se)` where `op(se)` is a (non-array) primop involving literals or constant variables (NEW)

I also included two unrelated fixes in separate commits.

Code coverage decreased slightly, from 82.89 to 82.65.


transform_constantfold.ml
```diff
@@ -73,17 +132,31 @@ let const_prop ({formals; instrs} : analysis_input) : instructions option =
...
+          in
+        fold (VarSet.fold try_prop (expr_vars exp) exp)
+      in
       match[@warning "-4"] instrs.(pc) with
-      | Decl_var (x, Simple (Constant l))
-        ->
-        VarMap.add x (Value l) cur
+      | Decl_var (x, e)
+      | Assign (x, e) ->
+        begin match (normalize e cur) with
```
gasche: This would be nicer and clearer with an `approx env e` auxiliary function that calls or replaces `normalize` and then converts to `Approx.t`.

transform_constantfold.ml
```diff
@@ -1,21 +1,89 @@
...
+  let module Approx = struct
+    type t = Unknown | Value of value
+    let equal a1 a2 = match a1, a2 with
+      | Unknown, Unknown -> true
+      | Unknown, Value _ | Value _, Unknown -> false
+      | Value v1, Value v2 -> Eval.value_eq v1 v2
+  end in
+
+  (* Constant fold an expression. *)
+  let fold exp =
```
gasche, line 28: proposal: just call `Eval.eval` instead, after checking that `exp_vars exp` is empty.

gasche (changes requested):

transform_constantfold.ml
```diff
@@ -26,36 +26,12 @@ let const_fold ({formals; instrs} : analysis_input) : instructions option =
...
-      | And -> (Simple (Constant (Bool (Eval.value_and v1 v2))))
-      | Or -> (Simple (Constant (Bool (Eval.value_or v1 v2))))
-      | (Neg | Not | Array_index | Array_length) -> assert false
-      end
-    | Op (_op, _es) -> exp
-    in
+    let main = {name=""; formals=[]; body=[]} in
+    let prog = {main; functions=[]} in
+    if VarSet.is_empty (expr_vars exp)
+    then Simple (Constant (Eval.eval prog Eval.Heap.empty Eval.Env.empty exp))
```
gasche: I removed the `prog` parameter in #126, so you'll have to rebase here, sorry.

gasche: Also I think the code would be more readable with the shorter case first (this is almost always the case): `if not (VarSet.is_empty (expr_vars exp))) ....`.

mhyee:
Rebased and all checks have passed.

gasche (approved):

PR #127, opened 2017-05-16
Generic mapper

gasche:
@mhyee I also went ahead and refactored part of your constant-folding work (it is your PR that gave me the idea of this), so I hope your tests are good!

@o- you won't see me using object-oriented very often, have a look!

### DIFF
--- a/edit.ml
+++ b/edit.ml
@@ -130,17 +130,10 @@ let freshen_assign ({instrs} as inp : analysis_input) (def : pc) =
     assert(false)
 open Instr
 
-let replace_var_in_exp var (exp : simple_expression) (in_exp : expression) : expression =
-  let in_simple_exp in_exp =
-    match in_exp with
-    | Constant _ -> in_exp
-    | Var x -> if x = var then exp else in_exp in
 
-  match in_exp with
-  | Simple se -> Simple (in_simple_exp se)
-  | Op (op, exps) ->
-    Op (op, List.map in_simple_exp exps)
-
-let replace_var_in_arg var (exp : simple_expression) (in_arg : argument) : argument =
-  replace_var_in_exp var exp in_arg
-
+let replace_var var simple_exp = object (self)
+  inherit Instr.map as super
+  method! simple_expression e = match e with
+    | Constant _ -> e
+    | Var x -> if x = var then simple_exp else e
+end
--- a/edit.ml
+++ b/edit.ml
@@ -110,62 +110,11 @@ let split_edge instrs pc label pc' =
   new_instrs.(pc_map pc) <- new_branch;
   new_instrs, pc_map
 
-let replace_uses_in_instruction old_name new_name instr : instruction =
-  let in_simple_expression (exp:simple_expression) : simple_expression =
-    match exp with
-    | Constant _ -> exp
-    | Var x -> if x = old_name then Var new_name else exp in
-  let in_expression exp : expression =
-    match exp with
-    | Simple se -> Simple (in_simple_expression se)
-    | Op (op, exps) ->
-      Op (op, List.map in_simple_expression exps) in
-  let in_arg e : argument = in_expression e in
-  let in_osr osr : osr_def =
-    match osr with
-    | Osr_var (x, exp) -> Osr_var (x, in_expression exp)
-  in
-  match instr with
-  | Call (x, f, exs) ->
-    assert(x <> old_name);   (* -> invalid scope *)
-    Call (x, in_expression f, List.map in_arg exs)
-  | Stop e ->
-    Stop (in_expression e)
-  | Return e ->
-    Return (in_expression e)
-  | Decl_var (x, exp) ->
-    assert(x <> old_name);   (* -> invalid scope *)
-    Decl_var (x, in_expression exp)
-  | Decl_array (x, def) ->
-    assert (x <> old_name);
-    let def = match def with
-      | Length e -> Length (in_expression e)
-      | List es -> List (List.map in_expression es)
-    in Decl_array (x, def)
-  | Assign (x, exp) ->
-    Assign (x, in_expression exp)
-  | Array_assign (x, index, exp) ->
-    let x' = if x = old_name then new_name else x in
-    Array_assign (x', in_expression index, in_expression exp)
-  | Drop x ->
-    if x = old_name then Drop new_name else instr
-  | Print exp ->
-    Print (in_expression exp)
-  | Assert exp ->
-    Assert (in_expression exp)
-  | Branch (exp, l1, l2) ->
-    Branch (in_expression exp, l1, l2)
-  | Osr {cond; target; map} ->
-    let cond = List.map in_expression cond in
-    let map = List.map in_osr map in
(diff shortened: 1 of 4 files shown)


o-:
> you won't see me using object-oriented very often, have a look!

haha, learn from the best. I'll look at it ;) need sleep first

Maybe transformations could be objects too? For example for debugging it would be very useful if a transformation knows its name and can log its changes...

edit.ml
```diff
@@ -110,61 +110,16 @@ let split_edge instrs pc label pc' =
...
-  | Label _ | Goto _ | Comment _ ->
-    assert (VarSet.is_empty (used_vars instr));
-    instr
+  let mapper = object (self)
+    inherit Instr.mapper as super
+    method! bound_variable x =
+      if x = old_name then new_name else x
+    method! binder x = x
+    method! instruction = function[@warning "-4"]
+      | Assign (x, e) -> Assign (x, self#expression e)
```
gasche: Note: this special-case behavior of *not* touching the assigned variable was kept from the previous version of the code, but I don't know if it's still necessary. (There is a call after this mapper is used that works on Assign instructions specifically, and it may have to be changed in synch.). It would be nicer (more regular) to get rid of it.

edit.ml
```diff
@@ -110,61 +110,16 @@ let split_edge instrs pc label pc' =
...
-  | Read x ->
-    assert (x <> old_name);
-    instr
-
-  | Label _ | Goto _ | Comment _ ->
-    assert (VarSet.is_empty (used_vars instr));
-    instr
+  let mapper = object (self)
+    inherit Instr.mapper as super
+    method! bound_variable x =
```
gasche: `method!` is like `method` but it explicitly indicates the intent to override an existing method. Otherwise it's easy to do a typo, and define a separate method instead of overriding the one we think. (It happened to me while writing this PR.)

edit.ml
mhyee: I agree, it would be nice to remove it. It's also a bit odd that `Assign` has a special case but `Array_assign` does not. But arrays were added later and maybe this case was missed?

It looks like this is being used in `freshen_assign`, which is used for hoisting assignments. But then we also removed const/mut, which probably affects the optimization.

@o-, I think you were the one who implemented the optimization? What do you think?

transform_constantfold.ml
```diff
@@ -24,73 +24,25 @@ let const_fold ({formals; instrs} : analysis_input) : instructions option =
...
-      (* Replace all expressions in the osr environment. *)
-      let map = List.map (fun (Osr_var (y, e)) -> Osr_var (y, replace e)) map in
-      let cond = List.map replace cond in
-      Osr {cond; target; map}
-    | Drop y
-    | Read y ->
-      instr
-    | Label _ | Goto _ | Comment _ -> instr in
+    method! expression exp =
+      let all_folded = ref true in
```
mhyee: This should technically be `all_propagated` (or something similar). I use "propagation" to mean "replacing constant variables with their values" and "folding" to mean "evaluating (closed) expressions."

edit.ml
```diff
@@ -110,61 +110,16 @@ let split_edge instrs pc label pc' =
   new_instrs, pc_map
 
 let replace_uses_in_instruction old_name new_name instr : instruction =
```
o-: can't we get rid of this api alltogether?

edit.ml
o-: well the method is called replace uses... `x <- 4` is not a use of x. `x[1] <- 4` is a use of x. I guess that's also why there was an assert in declare (which you removed) that the declared variable is not the one to rename.

more symmetric is not always more correct...

o-: To answer you question: yes the pass does the following:
```
...
x <- x
print x
```
--->
```
...
var x_1 <- x
print x_1 
```

Therefore obviously in the case:
```
...
x <- x
x <- x
print x
```
The correct transformation is
```
...
var x_1 = x
x <- x_1       # here the assignee has to stay untouched since it's a different use-def chain
print x
```

instr.ml
```diff
@@ -322,3 +322,82 @@ let checkpoint_label pc =
 
 let independent instr exp =
   VarSet.is_empty (VarSet.inter (changed_vars instr) (expr_vars exp))
+
+class mapper = object (m)
+  method variable x = x
```
o-: is this one really necessary. found it very confusing. just inline below. i think users should think about the difference. i don't see a reasonable justification for unifying the handling of the *lhs* of an assignment and a *variable* occuring in an expression (except that they happen to have the same type in our instrs).

instr.ml
```diff
@@ -322,3 +322,82 @@ let checkpoint_label pc =
 
 let independent instr exp =
   VarSet.is_empty (VarSet.inter (changed_vars instr) (expr_vars exp))
+
+class mapper = object (m)
```
o-: I found the naming a bit confusing.

1. I would call the thing a *Map* instead of a *Mapper*
2. If you really want *Map* as a type then it should be an interface (not sure if they exist in ocaml, no need to actually do this, since this is a simple case)
3. This class here is the *IdentityMap*

o- (commented):
I have some comments

edit.ml
gasche: by removing the optimization that relies on it, or do you have other plans?

instr.ml
gasche: Do you think that I could separate three/ways into binding, use, and assignment positions? If I formulated the mut-to-const optimization above as transforming uses but not assignments, would that be good with you?

o-: right, that might be more clear

edit.ml
o-: no, I thought of eg. returning the map object or even push down this method (since it has only one caller I think)

gasche: Returning the map object: sure, can do. Pushing down: I would propose to keep in Edit the transformations that are generic enough to be potentially of use to different analyses/transformations, and push down the optimization-specific ones. If I can reword this one to not have Assign a special case anymore, it becomes generic enough to stay -- otherwise it goes. On the other hand, `freshen_assign` should probably be pushed down in any case.

instr.ml
gasche: I'm ok with `map`, but I don't see the point in `identity_map`: sure, we don't do anything, but the point is to inherit it to define (non-identity) mappers -- so inheritance would immediately break the `identity` part. I think that `generic_map` would also work, but again why do long when we can do short? One approach would have been to define one `map` class for different types (so a mapper on instructions would be parametrized over or inherit a mapper on variables and on expressions), but here the idea is to define one class for all types at once (and we do finer distinctions than the type structure, although this could be fixed by adding more type synonyms).

o-:
btw. I am ok with the rest of the changes. you can regard my comments as suggestions and proceed as you see fit

gasche:
I just rebased the branch with the various changes suggested. I will try to merge it soon to avoid further conflicts -- it already conflicts with the `fixedness` branch that I will rebase after this one is merged.

mhyee (approved):

PR #128, opened 2017-05-17
[TODO] found some more brokenness, but don't have time to fix

o-:


gasche:
Continuing my crusade against big-step analyses, the next bug in the pipeline comes from the fact that hoist_assign is broken. The input program (simplified by your humble servant) is:

```
0|   var i
1|   i <- 0
2| loop1:
3|   branch (i != 10) body1 continue
4| body1:
5|   i <- (i + 1)
6| goto loop1
7|   continue:
8|   print i
```

and it gets optimized as follows by hoisting the `i  <- 0` write:

```
 var i_1 = 0
 var i = nil
loop1:
 branch (i_1 != 10) body1 continue
body1:
 goto loop1
continue:
 print i_1
```

Not very good, right? The loop termination test uses the hoisted constant `i_1`. The write after `body1` was rewritten into `i <- (i_1 + 1)` by the optimizer, but it is subsequently removed by an always-happy-to-please cleanup pass.

I don't see a very clean way to fix the way hoist_assign is currently implemented (as a big-step transformation). It is true that `i <- 0` write dominates all uses of `i `, so hoisting does not seem wrong. But it is wrong to replace the reads of `i` on lines 3, 5, 8 by `i_1`, given that the write to `i` on line 5 may invalidate that.

I tried to change `Edit.freshen_assign`, which rewrites all the uses of the hoisted variable, to say: don't rewrite a use if it has other pcs in its `reaching_defs` set (if it may see other writes to the same variable). But that results in a program where the original `i <- 0` write does not happen (it was hoisted and then nothing was replaced). So if one of the uses is *not* rewritten, then we should not remove the initial write (but turn it into a write of the hoisted assignment). Meh.

I will try that fix, but I'm tempted to go small-step again.

gasche:
The next bug is in hoisting drops. It occurs on code such as the following:

```
stop 0
goto label # unique jump to this label
label:
drop x
```

Right now the code decides to lift `drop x` over `label` (so just on the instruction over, as there is no other predecessor), it knows that it should jump over the `goto label` and does, and then it encouters `stop 0` and it fails on an assert that was put there in the code because, we reasoned when we wrote it, if you only go towards you predecessors you never encounter `stop` or `return`.

I need to make this code more robust by not assuming `pc - 1` and calling the predecessor. In passing I may also make it more robust to fallthrough labels (when the code flows from the instruction right above the label that is not a branch/goto), which would remove the need for the first fallthrough-removal pass.

o-:
superseeded by #135

PR #130, opened 2017-05-23
Fixedness

gasche:
I haven't fixed all the issues that @o- found in #128 yet, but this is a start.

I had to completely rewrite the "minimize lifetime" pass as it was broken in presence of multi-predecessor multi-successor edges (luckily, we have exactly one of those in the testsuite). See the commit message.

### DIFF
--- a/instr.ml
+++ b/instr.ml
@@ -129,6 +129,7 @@ type analysis_input = {
   formals : VarSet.t;
   instrs : instructions;
 }
+type analysis = analysis_input -> instructions option
 
 type heap_value =
   | Undefined
--- a/instr.ml
+++ b/instr.ml
@@ -125,11 +125,6 @@ type program = {
   main : afunction;
   functions : afunction list;
 }
-type analysis_input = {
-  formals : VarSet.t;
-  instrs : instructions;
-}
-type analysis = analysis_input -> instructions option
 
 type heap_value =
   | Undefined
--- a/transform_liveness.ml
+++ b/transform_liveness.ml
@@ -1,7 +1,8 @@
 open Instr
+open Types
 open Transform_utils
 
-let (insert_returns : analysis) = fun {formals; instrs} ->
+let (insert_returns : transform_instructions) = fun {formals; instrs} ->
   let stops = Analysis.stops instrs in
   let add_return pc =
     if not (List.mem pc stops) then Unchanged
@@ -16,7 +17,7 @@ let transform_with_prepass ~prepass transform = fun input ->
     | Some instrs -> {input with instrs} in
   transform input
 
-let add_drops : analysis =
+let add_drops : transform_instructions =
   let add_drops ({formals; instrs} as input) =
     (* assume [insert_returns] has run *)
     let scope = Scope.infer input in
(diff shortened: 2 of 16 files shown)


transform_liveness.ml
```diff
@@ -1,34 +1,46 @@
...
+    | _ -> InsertAfter [Return (Simple (Constant Nil))]
+  in Transform_utils.change_instrs add_return {formals; instrs}
+
+let transform_with_prepass ~prepass transform = fun input ->
+  let input = match prepass input with
+    | None -> input
+    | Some instrs -> {input with instrs} in
+  transform input
+
+let add_drops : analysis =
```
gasche: @aviralg this is the code I use to "add all drops" -- in this pass the idea is to first add all drops, and then hoist them as high as possible. We could easily apply that pass on all the codebase as a first preprocessing case. You would still have work to do as `return (x + y)` cannot drop `x, y`, and with your transformation into `result <- (x + y)` you can now drop them. But the only thing you have to do, after this pass here has run, is to compute the required_vars of the return expression, and drop those variables.

instr.ml
```diff
@@ -129,6 +129,7 @@ type analysis_input = {
   formals : VarSet.t;
   instrs : instructions;
 }
+type analysis = analysis_input -> instructions option
```
o-: i find `analysis` confusing. For me `analysis` is `analysis_input -> analysis_result` whereas this one is a transformation (ie. an analysis just looks at the code whereas a transformation creates new code).

transform_liveness.ml
```diff
@@ -1,34 +1,46 @@
...
-  let transform pc =
-    let keep_alive = keep_alive pc in
-    assert (pc < Array.length instrs);
-    let keep_alive_before = keep_alive_before pc in
-    let to_drop = VarSet.diff keep_alive_before keep_alive in
-    let drops = List.map (fun x -> Drop x) (VarSet.elements to_drop) in
-    if drops = []
-    then Unchanged
-    else Insert drops
+let (insert_returns : analysis) = fun {formals; instrs} ->
```
o-: why `analysis` instead of `transform_instructions`

o- (commented):
hi, so @mhyee  and I started looking at it. Thanks for working on this. I think we agree that the general approach is a very good idea. We did not have time to look at the changes to transform_livetime in detail. We'll try to find some more time to look at it.

transform_liveness.ml
gasche: Right now `transform_instructions` is defined in `Transform` which itself depends on all `Transform_*` modules, so that would create a cyclic dependency. We could have a `Transform_types` module that just export relevant types and type synonyms.

mhyee (changes requested):
I looked at the details of the transformation, and it looks good. It's pretty straightforward!

This should also fix many of the issues I've been running into (poor error messages, scoping issues) while implementing the vector example. Thanks!

I'm marking this as "request changes" because some of the tests are failing with Travis.

transform_liveness.ml
o-: I see. I wonder why it worked in the first place. Transform_types.ml sounds like a good idea.

gasche:
Thanks, I fixed the test failure.

transform_liveness.ml
gasche: I updated the pr with a `Types` module with non-Instr-specific type definitions.

mhyee (approved):

o-:
merci beaucoup

gasche:
Prego.

PR #132, opened 2017-05-28
Fixedness

gasche:
These additional fixes are fixing the `freshen_assign` part of the `hoist_assign` optimization along the lines of my description of the issue: https://github.com/reactorlabs/sourir/pull/128#issuecomment-304382851.

PR #135, opened 2017-05-31
extend the testrunner

o-:
* run tests with more inputs and all combination of passes
  (use make test_examples_long)
* run tests in parallel

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,9 @@ run: sourir
 test_examples: sourir
 	bash test_examples.sh
 
+test_examples_long: sourir
+	bash test_examples.sh --long
+
 clean:
 	ocamlbuild -clean
 
--- a/test_examples.sh
+++ b/test_examples.sh
@@ -1,11 +1,25 @@
 #!/usr/bin/env bash
-SOURIR="../sourir"
+
+LONG=$1
+
+export SOURIR="../sourir"
+
+function ncores {
+  if [[ "$OSTYPE" == "darwin"* ]]; then
+    echo `sysctl -n hw.ncpu || echo 8`
+  else
+    echo `nproc || echo 8`
+  fi
+}
+
+export OPTS="prune\nconst_fold\nhoist_assign\nhoist_drop\nmin_live"
+export INPUTS="0\n1\n3\nnil\ntrue\nfalse"
 
 # Move into examples directory
 pushd examples > /dev/null 2>&1
 
 # Create a temp directory
-TEMPDIR=`mktemp -d`
+export TEMPDIR=`mktemp -d`
 
 # Confirm that we created the directory
 if [[ ! "$TEMPDIR" || ! -d "$TEMPDIR" ]]; then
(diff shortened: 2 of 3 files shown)


gasche:
Does this supersede #128? I just restarted #128's CI test (it should pass now that #132 was merged) with the intent to eventually merge it.

o-:
yes it does. All checks seem green now

test_examples.sh
```diff
@@ -20,25 +34,86 @@ function cleanup {
...
+  echo $TEST >> $STATUS
+}
+
+export -f runtests runtest status
+
+export NUM_TESTS=`find . -name '*.sou' | wc -l`
+export STATUS="$TEMPDIR/status"
+touch $STATUS
+
+find . -name '*.sou' | xargs -n 1 -P `ncores` bash -c 'runtests $@'
```
gasche, line 114: Does `xargs -P n` properly handles races in writing to stdout/stderr, or do we risk garbled output if two tests are failing? I've used `parallel` to parallelize the OCaml testsuite, it properly groups output.

o-: it garbles the output. but then you just need to rerun one at a time to see it. I think it's a minor concern...

gasche: Currently running the tests takes 1.7s second on my machine when they work. Garbled output is going to make me spend more  than 1.7s second whenever it fails, and make me deal with a problem I don't want to have to deal with. Please try to use the `parallel` command instead (I think pretty much the same syntax should work; the manpage also has a "Function tester" example that describes this use-case), or maybe just drop the parallelization that we don't really need.

o-: oh, we need it. I run much more tests now. I try more inputs and in the `--long` mode I try all combinations of optimizations.

`parallel` is a dependency that you have to install (its not lsb) so it's an obstacle. Also afaik it does not exist on osx (I evaluated this choice in the past, the code is c/p from the rir testrunner).

I can make serial execution the default and parallel behind a flag. I really prefer faster tests over the output garbling. Here are the execution times on my machine:

```
# serial

[user@work sourir]$ time bash test_examples.sh --long
[45/46] ./array.sou sou 

real	0m45.547s
user	0m24.226s
sys	0m28.565s
[user@work sourir]$ time bash test_examples.sh
[45/46] ./array.sou sou 

real	0m3.541s
user	0m1.540s
sys	0m2.514s
```

```
# parallel

[user@work sourir]$ time bash test_examples.sh --long
[42/46] ./array.sou sou 

real	0m27.286s
user	0m34.210s
sys	0m39.616s
[user@work sourir]$ time bash test_examples.sh
[42/46] ./array.sou sou 

real	0m1.735s
user	0m2.010s
sys	0m3.065s
```

gasche: My (Fedora) machines does not have a `ncores` command (and I see no package of that name; how did you install it?). People can `brew install parallel`, and we can test `which parallel` first to default to the sequential implementation to make it an optional dependency.

Let's merge your version now, I'll send a `parallel` proposal as a PR on top of it.

o-: sure, I am just trying to not spend too much time on things that I consider not prio at the moment. this method I know works well.

`ncore` is a bash function (see higher up)

gasche: Not a priority? Output garbling is *the worst*, man.

o-: sorry :)

broken tranformations are strictly worse

Makefile
```diff
@@ -35,6 +35,9 @@ run: sourir
 test_examples: sourir
 	bash test_examples.sh
 
+test_examples_long: sourir
```
mhyee, line 38: Do we want `test_examples` or `test_examples_long` in the default make task?

gasche:
Let's make the short test the default, have `make test` run the long tests, and make sure Travis runs everything it can.

test_examples.sh
```diff
@@ -1,11 +1,25 @@
 #!/usr/bin/env bash
-SOURIR="../sourir"
+
+LONG=$1
+
+export SOURIR="../sourir"
+
+function ncores {
+  if [[ "$OSTYPE" == "darwin"* ]]; then
+    echo `sysctl -n hw.ncpu || echo 8`
```
mhyee, line 9: Is 8 a valid default choice? It seems that among the recent Intel processors, only the high performance ones come with "8" cores (4 physical cores with hyper-threading). "4" seems to be more common these days?

test_examples.sh
```diff
@@ -20,25 +34,86 @@ function cleanup {
 }
 trap cleanup EXIT
 
+export ALL_OPTS=$TEMPDIR/all_opts
+if [[ "$LONG" == "--long" ]]; then
+  p() { [ $# -eq 0 ] && echo || (shift; p "$@") | while read r ; do echo -e "$1,$r\n$r"; done }
```
mhyee, line 39: I know this is just a bash script for testing infrastructure, but a one-line comment explaining what this does would be nice.

(Also, I suppose running every possible *permutation* of optimizations isn't worth it.)

test_examples.sh
```diff
@@ -20,25 +34,86 @@ function cleanup {
...
+    echo " --------------------------------------------------"
+    exit 255
   fi
-done
+}
+
+function status {
+  name=$1
+  done=`wc -l < $STATUS`
+  echo -ne "\e[0K\r[${done}/${NUM_TESTS}] ${name} "
```
mhyee, line 92: This is more clever than necessary. After I run the tests, I see something like:

```
[      38/      46] ./sum.sou on_without_fallthrough.sou
```

Running tests in parallel screws up the counter, and only part of the line is cleared (`sum.sou` is not long enough to overwrite `regression_without_fallthrough.sou`.)

Also, the first time I ran this, I didn't even know all the tests had passed. I thought 38/46 meant there were some failures.

o-:
wow, I didn't think that this PR would gather comments :)

sorry, I am fighting with all the optimizer bugs right now. After that I can come back to it.

Makefile
o-: test_examples_long is too slow. It's fine since I made sure that travis runs it

test_examples.sh
o-: sure

test_examples.sh
o-: it generates all possible subsets. Not all permutations. This would be nice, but would take forever to run. This has helped quite a lot in finding bugs.

test_examples.sh
o-: Mhyeah, so the not being able to clear the line is a known bug of that script. I find it easy to live with honestly.

The counter error is a bug, basically at the end of a test a second `status` call should be added (now the status is only updated when a test starts). Concurrency is not the issue since they communicate through the append only status file.

mhyee: It's fine now that I know what's going on, but the first time, I thought we were running `sum.sou` and `on_without_fallthrough.sou` at the same time, so it was confusing.

I would prefer something clearer and less fancy, e.g. getting rid of the line clearing. In fact, I can live without the counter, especially since the tests (right now) are still very fast.

test_examples.sh
o-: haha, I looket at the parallel manpage. I am not gonna read that one :)

PR #136, opened 2017-05-31
fix hoist_assign

o-:
There where two things broken with the pass:

1. It might hoist assignments to a place where the rhs expression is invalid
2. It might destroy the scope

The first problem occurs for example in:

     var a = true
     branch (a==1) isint cont
    isint:
     a <- a+2
    cont:

If we hoist the assignment to line 2 the addition throws
a type error. The problem is solved by only hoisting assignments
with simple expressions rhs. A better approach would be nice...

The second problem arises for example in:

     branch ... a b
    a:
     var a = 1
     ...
     a <- 2
     ...
     drop a
     goto b

If we create a new variable a_1 inside the branch we need to
drop it before merging the scope with b. This problem is solved
by running a generic fix_scope pass which makes all missing drops
explicit.

### DIFF
--- a/transform_utils.ml
+++ b/transform_utils.ml
@@ -42,3 +42,45 @@ let change_instrs (transform : pc -> instruction_change) ({formals; instrs} : an
         acc_instr (pc+1) (instrs.(pc)::acc) changed
   in
   acc_instr 0 [] false
+
+(* This util can fix the scope after inserting a fresh variable in the graph.
+ * As an additional sideeffect it will make every drop explicit. *)
+let fix_scope : transform_instructions = fun {formals; instrs} ->
+  let merge pc cur incom =
+    let res = VarSet.inter cur incom in
+    if VarSet.equal res cur then None else Some res
+  in
+  let update pc cur =
+    let instr = instrs.(pc) in
+    let added = Instr.declared_vars instr in
+    let updated = VarSet.union cur added in
+    let dropped = Instr.dropped_vars instr in
+    VarSet.diff updated dropped
+  in
+  let initial_state = formals in
+  let scope = Analysis.forward_analysis initial_state instrs merge update in
+  let succs = Analysis.successors instrs in
+  let transform pc =
+    let succs = succs.(pc) in
+    assert(List.length succs <= 2);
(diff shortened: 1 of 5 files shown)


o-:
Our favorite footgun strikes again. Hoisting assignments breaks the scope (no surprise there). This time I took out the big hammer and implemented a fix_scope pass. This one first calculates the maximally possible scope (ie. the intersection of all incomming scopes) and then drops whatever is dangling.

This works obviously only on a split-edged graph. So it also does a split-edge prepass.

Maybe those two primitives could be used to simplify other passes we have?

o-:
Ugh, of course the first error is also present in the constant folding :(   (as our extensive test suite has shown). I'll work on that in a separate PR

transform_utils.ml
```diff
@@ -19,10 +19,76 @@ let change_instrs (transform : pc -> instruction_change) ({formals; instrs} : an
...
       | Unchanged ->
         acc_instr (pc+1) (instrs.(pc)::acc) changed
   in
   acc_instr 0 [] false
+
+
+(* Splits all edges that have multi-predecessors branching *)
+let normalize_graph instrs =
+  let rec normalize instrs preds pc =
+    let is_branch pred_pc = match[@warning "-4"] instrs.(pred_pc) with
```
mhyee: Formatting nitpick: put the match on a new line and indent it (and also indent the cases).

transform_utils.ml
```diff
@@ -19,10 +19,76 @@ let change_instrs (transform : pc -> instruction_change) ({formals; instrs} : an
       | Replace is ->
         acc_instr (pc+1) (List.rev_append is acc) true
       | Insert is ->
+        assert(match[@warning "-4"] instrs.(pc) with | Label _ -> false | _ -> true);
```
mhyee: Please add a comment explaining this insert. (The next time I see this code, I'll probably forget why there's an assert.)

mhyee:
What is the status of this?

o-:
the status is that I'd like to merge it :)

o-:
I'll go ahead and take silence as approval.

PR #138, opened 2017-05-31
Fix hoist drop

o-:
Hoist drop was broken in the case from examples/scope_err.sou

It was assumed that a split edge is only necessary if the first label
to encounter sees two branches and otherwise a pull is ok.

But e.g. in the case:

        a    b

       /  \  /

      c    d

we would happily assume at `d` that the graph is in split-edge form
(because only one branch is a successor). Then we would pull on `a`
and if that succeeds the push to `b` gets lost.

The workaround is to more strongly normalize the graph, ie. allow
mergeing control flow only through gotos. This fixes the missing push
problem. Secondly split edge is performed ahead of time to avoid having
to do non-local checks for all successors of predecessors of a label.

### DIFF
--- a/examples/scope_err.sou
+++ b/examples/scope_err.sou
@@ -1,0 +1,13 @@
+function main()
+stop 0
+
+function obj_check (var obj, var tpe)
+    var candidate
+    branch (obj == nil) error l1
+  l1:
+    candidate <- obj[0]
+    branch (candidate != tpe) error done
+  done:
+    return true
+  error:
+    return false
--- a/examples/type_tests.sou
+++ b/examples/type_tests.sou
@@ -1,0 +1,11 @@
+var a = false
+var b = false
+
+branch (a == 1) num cont
+
+num:
+# This assignment cannot be hoisted since it depends on the branch
+b <- (a+1)
+print b
+
+cont:
--- a/examples/type_tests2.sou
+++ b/examples/type_tests2.sou
@@ -1,0 +1,17 @@
+var a = false
+var b = false
+
+print a
+
+# This assignment can be hoisted since it does not depend on the branch
+b <- (a==true)
+
+goto bb
+bb:
+
+branch (a == 1) num cont
+
+num:
+print b
+
+cont:
--- a/tests.ml
+++ b/tests.ml
@@ -573,9 +573,11 @@ let do_test_codemotion = function () ->
   " in
   let expected = parse "
        goto bla
+      loop_1:
+       goto loop
       loop:
        x <- (x + y_1)
-       branch (x == 10) end loop
+       branch (x == 10) end loop_1
       end:
        drop x
        drop y
@@ -606,9 +608,12 @@ let do_test_codemotion = function () ->
        var y_1 = 1
        var x = 1
        var y = 2
+       goto loop
+      loop_1:
+       goto loop
       loop:
        x <- (x + y_1)
-       branch (x == 10) end loop
+       branch (x == 10) end loop_1
       end:
        drop x
        drop y
@@ -637,7 +642,7 @@ let do_test_codemotion = function () ->
   let expected = parse "
        var a = 1
        var b = 2
-       branch (a==b) ba bb
+       branch (a==b) ba bb_2
       ba:
        var d = 1
        var d_1 = d
@@ -652,6 +657,8 @@ let do_test_codemotion = function () ->
       bb_1:
        drop d
        drop d_1
+       goto bb
+      bb_2:
        goto bb
       bb:
        drop a
@@ -680,7 +687,7 @@ let do_test_codemotion = function () ->
   let expected = parse "
        var a = 1
        var b = 2
-       branch (a==b) ba bb
+       branch (a==b) ba bb_3
       ba:
        var d = 1
        var d_1 = d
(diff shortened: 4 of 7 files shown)


o-:
this is still broken. I think the original logic is flawed. I have to think more about it.

o-:
@gasche see the updated comment and second commit. We need an even stronger normalized graph and we can't do the split-edge lazily. Now it should work.

gasche:
I want to poke at your idea and try to contradict it. (I have a justification besides "I like to do that": maybe we need more poking and contradiction, at least this issue shows that my changes from last week could have benefitted from them.)

Do we really want to normalize eagerly? I unserstand that this example is problematic, but I have the impression that the following strategy would also solve it (without normalization): split branch predecessors unless there is a single predecessor, and then if it's a branch pull it.

I'm not opposed to normalization in general, but since two weeks ago I'm opposed to correctness dependencies between passes. If we want to enforce a certain graph structure, I would propose to enforce this at the Check level directly, and be able to assume (and check) that it holds everywhere and is always preserved. If we are not willing to do this, then I think being lazy and defensive is the most robust choice.

(I still have in mind that we could move to an explicit control-flow-graph representation. It probably wouldn't have helped directly here (although a clearer reasoning on control flow could have helped us think about it), but it would certainly have helped with the fallthrough label business of my last bugfix.)

gasche:
I want to poke at your idea and try to contradict it. (I have a justification besides "I like to do that": maybe we need more poking and contradiction, at least this issue shows that my changes from last week could have benefitted from them.)

Do we really want to normalize eagerly? I unserstand that this example is problematic, but I have the impression that the following strategy would also solve it (without normalization):
- split branch predecessors unless there is a single predecessor
- if there is a single branch predecessor, pull it
- otherwise all goto/fallthrough predecessors, push above

I'm not opposed to normalization in general, but since two weeks ago I'm opposed to correctness dependencies between passes. If we want to enforce a certain graph structure, I would propose to enforce this at the Check level directly, and be able to assume (and check) that it holds everywhere and is always preserved. If we are not willing to do this, then I think being lazy and defensive is the most robust choice.

(I still have in mind that we could move to an explicit control-flow-graph representation. It probably wouldn't have helped directly here (although a clearer reasoning on control flow could have helped us think about it), but it would certainly have helped with the fallthrough label business of my last bugfix.)

o-:
> but I have the impression that the following strategy would also solve it (without normalization): 

That's the whole problem, it doesn't.

> split branch predecessors unless there is a single predecessor, and then if it's a branch pull it.

The problem is with your last part. *if it's a branch pull it*. That does not rule out that the other target of the branch is a multi predecessor node. Our pulling will just forget the other predecessors.

Ie. in my above example you would see c has a single predecessor. Alas you pull on `a`, which means you move instructions from `c` and `d` to `a`. But the pulling would first need to do a push to `b`. This is a non-trivial fix since the pulling logic would need to also do pushes (and it might cause a ripple effect along the chain...)

o-:
Why are you opposed to normalizing eagerly? Almost all of our painful bugs come from the fact that our IR is not structured enough regarding control flow and scoping. Let's tighten it a bit, please. I am ok with preprocessing early and enforcing it at check time.

gasche:
For the sake of the argument: our "pull" code could check that the two successors are good, and block otherwise (asking to pull on one non-good predecessor). This would not loop, given that pushing this non-good predecessor would split the edge, decreasing the number of non-good predecessors for the branch.

gasche:
> Why are you opposed to normalizing eagerly?

*Opposed* would be a strong world. But my impression (to explain the bug plague we had) is that when the invariants we are interested in are neither reflected in the syntax nor at least enforced by the pieces of code we spend our days looking at (the tests, the examples, etc.), then it's easy to forget about them when we reason about our programs.

gasche:
Here would be an idea to have the invariant we want enforced in the syntax: give each branch instructions a (unique) *name* (`branch<foo> cond l1 l2`), and distinguish "goto label", which are as today, and "branch labels", that start with the branch name:

```
while:
branch<loop> cond body end
loop.body:
  ...
  goto while
loop.end:
  ...
```

My previous ideas (I think in the very first mail suggesting the multi-predecessor multi-successor issue) was to just have distinct namespaces for `goto labels` and `branch labels`, but that does not enforce single-use of branch labels by itself. Remembering the single-use restriction is not natural; combining this with a naming discipline (that we know must be conflict-free) makes it more natural.

o-:
ok, what I am currently working on is a cfg representation that enforces this structure. I think changing the instructions representation at this point is to hard. but we can easily enforce the cfg to be well-formed after every pass.

o-:
@gasche I think we should agree on a strategy to deal with it soon. I tried to enforce the good properties by constructing a cfg in #141 . But it's kind of painful and you have to make sure to fix the graph after every step.

I still believe that this would be the least painful solution. But it's more coding than I hoped for.

I am a bit hesitant to add it to the syntax. It feels strange.

o-:
@gasche you convinced me that a syntactical approach makes sense. I didn't go all the way of having named branches. I believe it is enough to syntactically distinguish between branch and goto targets. This allows for lightweight checking of the graph structure. In you case of the named branches you would still need to check that the names are unique, so not much is lost imho.

Anyways, please have a look at #144 for a patch that superseeds this approach.

PR #140, opened 2017-05-31
update the scope definition

o-:
as recently discussed it is no longer considered valid if a variable
has two declaration.

This change updates the scope checker to reflect this decision.

### DIFF
--- a/scope.ml
+++ b/scope.ml
@@ -43,15 +43,15 @@ let infer ({formals; instrs} : analysis_input) : inferred_scope array =
     in
     let update pc cur =
       let instr = instructions.(pc) in
-      let declared = Instr.declared_vars instr in
-      let declared' = VarSet.elements declared in
-      let declared' = List.map (fun v -> (v, pc)) declared' in
-      let declared' = DeclSet.of_list declared' in
+      let declared_vars = Instr.declared_vars instr in
+      let decls = VarSet.elements declared_vars
+                |> List.map (fun v -> (v, pc))
+                |> DeclSet.of_list in
       let info = cur.info in
-      let shadowed = VarSet.inter (decl_as_var_set info) declared in
+      let shadowed = VarSet.inter (decl_as_var_set info) declared_vars in
       if not (VarSet.is_empty shadowed) then
         raise (DuplicateVariable (shadowed, pc));
-      let updated = DeclSet.union info declared' in
+      let updated = DeclSet.union info decls in
       let dropped = VarSet.elements (Instr.dropped_vars instr) in
       let remove set el = DeclSet.filter (fun (v, _) -> v <> el) set in
       let final_info = List.fold_left remove updated dropped in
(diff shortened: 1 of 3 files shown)


scope.ml
```diff
@@ -24,24 +36,29 @@ let infer ({formals; instrs} : analysis_input) : inferred_scope array =
...
-      if not (VarSet.equal cur.info incom.info)
+      if not (DeclSet.equal cur.info incom.info)
       then raise (IncompatibleScope (cur, incom, pc))
       else if PcSet.equal cur.sources incom.sources then None
       else Some { info = cur.info; sources = PcSet.union cur.sources incom.sources }
     in
     let update pc cur =
       let instr = instructions.(pc) in
-      let added = Instr.declared_vars instr in
+      let declared = Instr.declared_vars instr in
```
mhyee: Could you use different names for `declared` and `declared'`? The names are too similar for me.

scope.ml
```diff
@@ -24,24 +36,29 @@ let infer ({formals; instrs} : analysis_input) : inferred_scope array =
...
+      let declared' = VarSet.elements declared in
```
mhyee: Would this be clearer if it was written as
```ocaml
let declared' = VarSet.elements declared
             |> List.map (fun v -> (v, pc))
             |> DeclSet.of_list
```

scope.ml
```diff
@@ -24,24 +36,29 @@ let infer ({formals; instrs} : analysis_input) : inferred_scope array =
...
+      let declared' = List.map (fun v -> (v, pc)) declared' in
+      let declared' = DeclSet.of_list declared' in
       let info = cur.info in
-      let shadowed = VarSet.inter info added in
+      let shadowed = VarSet.inter (decl_as_var_set info) declared in
```
mhyee: It seems this is the only other place where `declared` is used. Maybe you could write this as
```ocaml
let shadowed = decl_as_var_set (DeclSet.inter info declared')
```
and then get rid of `declared`?

mhyee (approved):

scope.ml
```diff
@@ -9,7 +9,19 @@ exception UndeclaredVariable of VarSet.t * pc
 exception ExtraneousVariable of VarSet.t * pc
 exception DuplicateVariable of VarSet.t * pc
 
-type scope_info = VarSet.t
+module Declaration = struct
+  type t = string * pc
+  let compare a b =
```
gasche, line 14: `Pervasives.compare` would correctly compare these pairs

gasche:
Why are you using a `(string * pc)` list instead of a map from variable declarations to pcs?

o-:
because I don't understand how maps work in ocaml...

gasche:
`module KSet = Set.Make(K)` creates a set module over keys of type `K.t`, where the sets have the type `KSet.t`. `module KMap = Map.Make(K)` creates a map module over keys of type `K.t`, where maps from keys to value of type `foo` have type `foo KMap.t`. If you are comfortable with sets, you should be comfortable with maps as well.

PR #142, opened 2017-06-01
combine checkpoint label with osr instruction

o-:
Checkpoint labels are a bit of a hassle when it comes to transformations.
They have to be threated differently from the other labels. For example
they have to be preserved even if there is no goto/branch using them.

Also if we want to enforce some structure on the cfg in the future they
will present an exception again.

Therefore I combined them into the osr instruction.

### DIFF
--- a/parser.messages
+++ b/parser.messages
@@ -1,6 +1,6 @@
 program: BRANCH NIL IDENTIFIER TRIPLE_DOT 
 ##
-## Ends in an error in state: 121.
+## Ends in an error in state: 126.
 ##
 ## instruction -> BRANCH expression label . label [ NEWLINE ]
 ##
@@ -15,7 +15,7 @@ instruction
 
 program: BRANCH NIL TRIPLE_DOT 
 ##
-## Ends in an error in state: 120.
+## Ends in an error in state: 125.
 ##
 ## instruction -> BRANCH expression . label label [ NEWLINE ]
 ##
@@ -29,7 +29,7 @@ example "foo", is now expected to construct a branch instruction
 
 program: BRANCH TRIPLE_DOT 
 ##
-## Ends in an error in state: 119.
+## Ends in an error in state: 124.
 ##
 ## instruction -> BRANCH . expression label label [ NEWLINE ]
 ##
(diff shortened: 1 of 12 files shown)


parser.messages
```diff
@@ -57,9 +57,10 @@ example "(x + 1)", is now expected to construct a variable declaration
 
 program: VAR IDENTIFIER TRIPLE_DOT 
 ##
-## Ends in an error in state: 99.
+## Ends in an error in state: 31.
 ##
-## instruction -> VAR variable . EQUAL expression [ NEWLINE ]
+## var_def -> VAR variable . EQUAL expression [ RBRACKET NEWLINE COMMA ]
+## var_def -> VAR variable . [ RBRACKET NEWLINE COMMA ]
```
mhyee, line 63: Shouldn't the error message be updated for the new production rule that was added?

parser.messages
```diff
@@ -71,9 +72,10 @@ Parsing an instruction, we parsed "var <var>" so far; the equal sign
 
 program: VAR TRIPLE_DOT 
 ##
-## Ends in an error in state: 98.
+## Ends in an error in state: 30.
 ##
-## instruction -> VAR . variable EQUAL expression [ NEWLINE ]
+## var_def -> VAR . variable EQUAL expression [ RBRACKET NEWLINE COMMA ]
+## var_def -> VAR . variable [ RBRACKET NEWLINE COMMA ]
```
mhyee, line 78: Same; another production rule was added here.

parser.messages
```diff
@@ -348,6 +350,7 @@ program: PRINT LPAREN TRIPLE_DOT
 ## Ends in an error in state: 36.
 ##
 ## expression -> LPAREN . simple_expression infixop simple_expression RPAREN [ RPAREN RBRACKET NEWLINE LPAREN IDENTIFIER COMMA ]
+## expression -> LPAREN . prefixop simple_expression RPAREN [ RPAREN RBRACKET NEWLINE LPAREN IDENTIFIER COMMA ]
```
mhyee, line 353: The error message needs to be updated for this rule.

mhyee (commented):
I just remembered that the semantics file is also slightly behind; `osr` now takes a list of expressions instead of a single expression.

PR #144, opened 2017-06-02
Graph discipline2

o-:
This change enforces a more rigorous graph discipline. There is a distinction
between merge labels and branch labels. Merge labels can have an arbitrary
number of incomming 'gotos'. Branch labels can have exactly one incomming
source from a branch instruction.

To make checking the graph a local property branch labels must begin with '$'.

The graph discipline is strictly enforced in the check module.

In theory merge labels would not have a problem with fallthrough controlflow.
But to make the structure uniform fallthrough is forbidden too.

The change simplifies reasoning about instruction movements as can be seen
in the hoist_drop pass.

depends on #142  and obsoletes #138

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-OCAMLBUILD=ocamlbuild
+OCAMLBUILD=ocamlbuild -cflag -g
 
 # use the ocamlfind library manager
 OCAMLBUILD+= -use-ocamlfind
--- a/check.ml
+++ b/check.ml
@@ -17,7 +17,12 @@ exception FunctionDoesNotExist of identifier
 exception VersionDoesNotExist of identifier * label
 exception InvalidNumArgs of pc
 exception InvalidArgument of pc * argument
+exception BranchLabelReused of pc
+exception DuplicateLabel of label
 exception MissingReturn
+
+exception FallthroughLabel of pc
+exception EntryPointIsLabel
 
 module IdentifierSet = Set.Make(Identifier)
 
@@ -66,6 +71,7 @@ let well_formed prog =
       then raise (InvalidNumArgs pc)
     in
 
+    let preds = Analysis.predecessors instrs in
     let check_fun_ref instr =
       let rec check_value = function
         | Nil | Bool _ | Int _ | Array _ -> ()
@@ -90,15 +96,15 @@ let well_formed prog =
       | Decl_var (_, e)
       | Decl_array (_, Length e)
       | Assign (_, e)
-      | Branch (e, _, _)
       | Print e
       | Assert e
       | Stop e
       | Return e
+      | Branch (e, _, _)
         -> check_expr e
       | Decl_array (_, List es) -> List.iter check_expr es
       | Drop _ | Read _
-      | Label _ | Goto _ | Comment _ -> ()
+      | Goto _ | Label _ | Comment _ -> ()
       | Array_assign (_, i, e) ->
         check_expr i;
         check_expr e;
@@ -106,6 +112,8 @@ let well_formed prog =
         List.iter check_expr cond;
         List.iter check_osr map
     in
+
+    let seen = ref [] in
 
     (* Check correctness of calls and osrs *)
     let check_instr pc instr =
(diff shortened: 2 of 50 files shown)


Makefile
```diff
@@ -1,4 +1,4 @@
-OCAMLBUILD=ocamlbuild
+OCAMLBUILD=ocamlbuild -cflag -g
```
gasche, line 1: this should not be necessary because the `true: debug` line in `_tags` should have the same effect. Why did you feel the need for it?

gasche:
I like the idea.

In terms of the implementation I wonder if it would not be nicer to separate the two `label` instructions, because we could have

```ocaml
type merge_label = ML of raw_label
type branch_label = BL of raw_label

type instr =
| ...
| Merge_label of merge_label
| Branch_label of branch_label
| Goto of merge_label
| Branch of expr * branch_label * branch_label
| Osr of { label : merge_label; ... }
| ...
```

and you would end up with substantially less `label -> label_type` conversion noise.

(Of course we could consider merging what you have now, and making that transformation afterwards.)

Makefile
o-: ups, that was not meant to be in there. I still wonder if ocamldebug is really hideous or if I don't know how to use it.

o-:
> In terms of the implementation I wonder if it would not be nicer to separate the two label instructions, because we could have

I initially wanted to go this way. The reason I went with this version was that I wanted to keep the 1:1 mapping between our source language instruction and instruction type.

Update: also there are a number of places where you do not care about the kind of label and there it would introduce a lot of noise. I suspect you would need to extend about 20 pattern matches that currently just match `| Label _`.

Basically either one is fine. If you find two instructions much better I can also still change it.

gasche:
I propose to merge this now, and consider the change as a later PR, so that the diff gets evaluated on its own effect instead of mixing the current restructuring and the change-of-representation diffs.

Makefile
mhyee: Reminder: this PR got merged before the debug flags were removed.

check.ml
```diff
@@ -127,6 +135,30 @@ let well_formed prog =
...
+        let _ = Instr.resolve instrs (MergeLabel l) in
+        check_fun_ref instr
+      | Branch (e, l1, l2) ->
+        let t1 = Instr.resolve instrs (BranchLabel l1) in
+        if (preds.(t1) <> [pc]) then raise (BranchLabelReused t1);
+        let t2 = Instr.resolve instrs (BranchLabel l2) in
+        if (preds.(t2) <> [pc]) then raise (BranchLabelReused t2);
+        if (t1 = t2) then raise (BranchLabelReused t1);
+        check_fun_ref instr
+      | Label (MergeLabel l| BranchLabel l) ->
```
mhyee, line 148: Nitpick: missing space after `MergeLabel l`.

check.ml
```diff
@@ -127,6 +135,30 @@ let well_formed prog =
...
+        if List.mem l !seen then raise (DuplicateLabel l);
+        seen := l :: !seen;
+        (* Entry point cannot be a label because if it where a branch
```
mhyee, line 151: Typo: where -> were

check.ml
```diff
@@ -127,6 +135,30 @@ let well_formed prog =
...
+         * target it would have two incomming control flows. *)
```
mhyee, line 152: Typo: incomming -> incoming

disasm.ml
```diff
@@ -58,8 +58,9 @@ let disassemble_instrs buf ?(format_pc = no_line_number) (prog : instructions) =
     | Drop var                        -> pr buf " drop %s" var
     | Assign (var, exp)               -> pr buf " %s <- %a" var dump_expr exp
     | Array_assign (var, index, exp)  -> pr buf " %s[%a] <- %a" var dump_expr index dump_expr exp
-    | Branch (exp, l1, l2)            -> pr buf " branch %a %s %s" dump_expr exp l1 l2
-    | Label label                     -> pr buf "%s:" label
+    | Branch (exp, l1, l2)            -> pr buf " branch %a $%s $%s" dump_expr exp l1 l2
+    | Label (MergeLabel label)        -> pr buf "%s:" label
+    | Label (BranchLabel label)        -> pr buf "$%s:" label
```
mhyee, line 63: Nitpick: extra space screws up the alignment.

instr.ml
```diff
@@ -55,13 +55,16 @@ and instruction =
...
   | Branch of expression * label * label
-  | Label of label
+  | Label of label_type
   | Goto of label
   | Print of expression
   | Assert of expression
   | Stop of expression
   | Osr of {label : label; cond : expression list; target : label position; map : osr_def list; }
   | Comment of string
+and label_type =
```
mhyee: These should probably be `Merge_label` and `Branch_label`.

mhyee (commented):
I'll open a new issue so we can discuss splitting the label instruction.

(As a side note, I find

```
goto loop_body
$loop:
goto loop_body
loop_body:
```
a lot messier, which is a shame. But we've gone through enough pain from reusing labels.)

PR #145, opened 2017-06-02
Autofix

o-:
on top of #144 a preprocessing pass to automatically fix the graph. Removes fallthrough and makes branch targets unique.

### DIFF
--- a/transform_fix.ml
+++ b/transform_fix.ml
@@ -4,8 +4,8 @@ open Types
 
 let remove_falltrough ({instrs} as inp : analysis_input) =
   let normalize pc =
-    let fresh_label label tag = (Edit.fresh_label instrs label) ^ "_" ^ tag in
-    let has_fallthrough label pc =
+    let fresh_label label tag = (Edit.fresh_label instrs (label ^ "_" ^ tag)) in
+    let has_fallthrough pc =
       match instrs.(pc-1) with
       | Decl_var _ | Decl_array _
       | Assign _ | Array_assign _
@@ -15,13 +15,15 @@ let remove_falltrough ({instrs} as inp : analysis_input) =
     in
     match[@warning "-4"] instrs.(pc) with
     | Label (MergeLabel l) ->
-        if has_fallthrough l pc
+        if has_fallthrough pc
         then InsertBeforeLabel [ Goto l ]
         else Unchanged
     | Label (BranchLabel l) ->
         let l' = fresh_label l (string_of_int pc) in
-        if has_fallthrough l pc
-        then ReplaceLabel [ Goto l'; Label (BranchLabel l); Goto l'; Label (MergeLabel l') ]
+        if has_fallthrough pc
+        then ReplaceLabel [ Goto l';
+                            Label (BranchLabel l); Goto l';
+                            Label (MergeLabel l') ]
         else Unchanged
     | _ -> Unchanged
   in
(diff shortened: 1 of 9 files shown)


transform_utils.ml
```diff
@@ -3,22 +3,36 @@ open Types
...
-      | Insert is ->
+      | InsertBefore is ->
+        (* Inserting before a label is undefined. There is no sensible definition of what
+         * position is 'before'. *)
+        assert(not (is_label pc));
+        acc_instr (pc+1) (instrs.(pc) :: List.rev_append is acc) true
+      | ReplaceLabel is ->
+        assert(is_label pc);
+        acc_instr (pc+1) (List.rev_append is acc) true
+      | InsertBeforeLabel is ->
```
mhyee, line 34: Please put a comment saying that this is for inserting `goto`s to fix fallthroughs. Otherwise, it seems odd because a few lines above, a comment warns about the dangers of inserting before a label!

transform_fix.ml
```diff
@@ -0,0 +1,72 @@
+open Instr
+open Transform_utils
+open Types
+
+let remove_falltrough ({instrs} as inp : analysis_input) =
+  let normalize pc =
+    let fresh_label label tag = (Edit.fresh_label instrs label) ^ "_" ^ tag in
+    let has_fallthrough label pc =
```
mhyee: The `label` parameter seems to be unused?

transform_fix.ml
```diff
@@ -0,0 +1,72 @@
...
+        if has_fallthrough l pc
+        then ReplaceLabel [ Goto l'; Label (BranchLabel l); Goto l'; Label (MergeLabel l') ]
+        else Unchanged
+    | _ -> Unchanged
+  in
+  change_instrs normalize inp
+
+let make_branch_targets_unique ({instrs} as inp : analysis_input) =
+  let normalize pc =
+    let incomming_branches label =
```
mhyee: Typo: incomming -> incoming

(And also in other places.)

mhyee (commented):
For now, I think it's nice that we have this feature behind a flag, and that it isn't enabled by default. I don't think we should use this as a crutch, because I worry we'll forget about the invariants.

Also, please add some tests for this. (In the test suite and not examples. I think all our examples should be well-formed.)

o-:
> Also, please add some tests for this.

mhyeah I am not really happy with our testsuite and was reluctant to add even more tests for a feature that is essentially not in our tcb. They are way to brittle and fixing the testsuite has become a huge part of making bigger changes (which is exactly the opposite of their supposed reason d'etre). The problem is that the main testing primitives we use are to coarse grained (ie. they compare a whole function) such that unrelated changes will break all the tests...

But fine I added some now...

o-:
then again, adding the tests did reveal some problems  -- thanks for your insistence :)

transform_fix.ml
```diff
@@ -15,44 +15,54 @@ let remove_falltrough ({instrs} as inp : analysis_input) =
...
     | Branch (e, l1, l2) ->
-        let i1, i2 = incomming_branches l1, incomming_branches l2 in
+        let i1, i2 = incoming_branches l1, incoming_branches l2 in
         begin match i1, i2 with
         | _, [] -> assert(false)
         | [], _ -> assert(false)
         | [pc1], [pc2] ->
             assert(pc1 = pc && pc2 = pc);
-            Unchanged
+            if pc1 <> pc2 then Unchanged else
```
mhyee: Because of the assert, isn't it guaranteed that `pc1` and `pc2` are equal? So either the condition is wrong or the assert should be removed.

o-: good catch

o-: should have been `l1<>l2`. It's fixed now

transform_fix.ml
```diff
@@ -15,44 +15,54 @@ let remove_falltrough ({instrs} as inp : analysis_input) =
...
     | Branch (e, l1, l2) ->
-        let i1, i2 = incomming_branches l1, incomming_branches l2 in
+        let i1, i2 = incoming_branches l1, incoming_branches l2 in
         begin match i1, i2 with
         | _, [] -> assert(false)
         | [], _ -> assert(false)
         | [pc1], [pc2] ->
             assert(pc1 = pc && pc2 = pc);
-            Unchanged
+            if l1 <> l2 then Unchanged else
```
mhyee: Still a bit confused about this, sorry.

If `incoming_branches` returned singleton lists, then `l1` and `l2` have to be different. If `l1` and `l2` are the same, then because of lines 39-40 above, `incoming_branches` would return a list with at least two `pc`s.

o-: Obviously I am the one who is confused :) Of course you are right.

o-: rinse and repeat...

mhyee (approved):
Looks good now!

Of course, my comments have delayed the merge and now you have to resolve merge conflicts...

o-:
mhm, and I have inherited the new bugs

o-:
This is now blocked. #149 needs to land first...

PR #146, opened 2017-06-02
Function Inlining

aviralg:
An implementation of function inlining which inlines functions in the program until it encounters recursive calls. OSR is not handled correctly.

### DIFF
--- a/analysis.ml
+++ b/analysis.ml
@@ -180,12 +180,15 @@ end
 
 exception DuplicateFormalParameter
 
-let as_var_set (formals : formal_parameter list) =
+let as_var_list (formals : formal_parameter list) =
   let to_var (Param x) = x in
-  let formals' = VarSet.of_list (List.map to_var formals) in
-  if (List.length formals) <> (List.length (VarSet.elements formals')) then
+  let formals' = List.map to_var formals in
+  if (List.length formals) <> (VarSet.cardinal (VarSet.of_list formals')) then
     raise DuplicateFormalParameter;
   formals'
+
+let as_var_set (formals : formal_parameter list) =
+  VarSet.of_list (as_var_list formals)
 
 let as_var_map formals =
   let formals = VarSet.elements formals in
--- a/edit.ml
+++ b/edit.ml
@@ -1,37 +1,36 @@
 open Instr
 open Types
 
-(* TODO unify those 3 functions *)
-let fresh_var instrs var =
-  let cand i = var ^ "_" ^ (string_of_int i) in
-  let is_fresh cand_var instr =
-    let existing = declared_vars instr in
-    not (VarSet.mem cand_var existing) in
+module StringSet = Set.Make(String)
+
+let fresh (existing : StringSet.t) (name : string) : string =
+  let cand i = name ^ "_" ^ (string_of_int i) in
+  let is_fresh cand_var = not (StringSet.mem cand_var existing) in
   let rec find i =
     let cand_var = cand i in
-    if Array.for_all (is_fresh cand_var) instrs then cand_var else find (i+1) in
+    if is_fresh cand_var then cand_var else find (i+1) in
   find 1
 
-let fresh_label instrs label =
-  let cand i = label ^ "_" ^ (string_of_int i) in
-  let is_fresh cand_lab instr = match[@warning "-4"] instr with
-    | Label (BranchLabel l) -> l <> cand_lab
-    | Label (MergeLabel l) -> l <> cand_lab
-    | _ -> true in
-  let rec find i =
-    let cand_lab = cand i in
-    if Array.for_all (is_fresh cand_lab) instrs
-    then cand_lab else find (i+1) in
-  find 1
+let fresh_var instrs var =
+  let existing = Array.fold_left
+      (fun acc instr -> VarSet.union (declared_vars instr) acc)
+      VarSet.empty instrs
+  in
+  fresh existing var
+
+let extract_labels instrs =
+  let add_label acc instr =
+    match[@warning "-4"] instr with
+    | Label (MergeLabel l | BranchLabel l) -> LabelSet.add l acc
+    | _ -> acc
+  in
+  Array.fold_left add_label LabelSet.empty instrs
+
+let fresh_label instrs label = fresh (extract_labels instrs) label
 
 let fresh_version_label (func : afunction) label =
-  let cand i = label ^ "_" ^ (string_of_int i) in
   let existing = List.map (fun {label} -> label) func.body in
-  let rec find i =
-    let cand_lab = cand i in
-    if not (List.mem cand_lab existing)
-    then cand_lab else find (i+1) in
-  find 1
+  fresh (StringSet.of_list existing) label
 
 type pc_map = int -> int
 
@@ -79,4 +78,12 @@ let replace_var var simple_exp = object (self)
   method! simple_expression e = match e with
     | Constant _ -> e
     | Var x -> if x = var then simple_exp else e
+end
+
+let replace_all_vars replacements = object (self)
+  inherit Instr.map as super
+  method replace x = try List.assoc x replacements with Not_found -> x
+  method! binder x = self#replace x
+  method! variable_assign x = self#replace x
+  method! variable_use x = self#replace x
 end
--- a/examples/inline_advanced.sou
+++ b/examples/inline_advanced.sou
@@ -1,0 +1,25 @@
+function main ()
+ var x = 24
+ call y1 = 'foo (x)
+ assert (y1 == 24)
+ call y2 = 'foo (x)
+ assert (y2 == 24)
+ call y3 = 'foo (x)
+ assert (y3 == 24)
+
+function foo (var x)
+ var m = x
+ call y1 = 'bar (m)
+ assert (y1 == 24)
+ call y2 = 'bar (y1)
+ assert (y2 == 24)
+ return y2
+
+function bar (var x)
+   branch (x == 10) $continue $done
+$continue:
+  var x1 = (x - 1)
+  call x2 = 'bar (x1)
+  return x2
+$done:
+  return x
--- a/instr.ml
+++ b/instr.ml
@@ -1,12 +1,6 @@
-module Variable = struct
-  type t = string
-  let compare = String.compare
-end
+module Variable = String
 
-module Label = struct
-  type t = string
-  let compare = String.compare
-end
+module Label = String
 
 module Address : sig
   type t = private int
@@ -23,6 +17,8 @@ type variable = Variable.t
 type label = Label.t
 
 module VarSet = Set.Make(Variable)
+
+module LabelSet = Set.Make(Label)
 
 module Identifier = struct
   type t = string
@@ -305,6 +301,17 @@ let lookup_fun (prog : program) (f : identifier) : afunction =
   if f = "main" then prog.main else
   try List.find (fun {name} -> name = f) prog.functions with
   | Not_found -> raise (FunctionDoesNotExist f)
+
+let replace_fun ({main; functions} as prog : program) (func : afunction) : program =
+  let rec replace_fun' rest =
+    match rest with
+    | [] -> []
+    | func' :: rest' ->
+      if func'.name = func.name then func :: rest'
+      else func' :: replace_fun' rest'
+  in
+  if (func.name = main.name) then {prog with main = func}
+  else {prog with functions = replace_fun' functions}
 
 let get_version (func : afunction) (v : label) : version =
   List.find (fun {label} -> label = v) func.body
--- a/transform_inline.ml
+++ b/transform_inline.ml
@@ -1,0 +1,261 @@
+open Instr
+open Edit
+
+(* Unlike `analysis_input`, this preserves the order of `formals`. This
+   facilitates matching the formals to the arguments passed at a callsite.*)
+type inlining_input = {
+  formals : variable list;
+  instrs : instructions;
+}
+
+let as_inlining_input (func : afunction) (version : version) : inlining_input =
+  { formals = Analysis.as_var_list func.formals; instrs = version.instrs }
+
+(* FUNCTION INLINING *)
+(* Given a program, inline the functions in it to the maximum possible extent.
+   Transitively recursive calls are inlined until they become self recursive.
+   Self recursive callsites are not inlined. Intermediate inlined versions are
+   stored in the caller. *)
+let inline ({main; functions} as orig_prog : program) : program option =
+
+  (* Given the formals and instructions of a function, get all the declared and
+     used vars in the function. *)
+  let function_vars {formals; instrs} =
+    let all_declared_vars instrs = Array.fold_left
+        (fun vars instr -> VarSet.union vars (declared_vars instr))
+        VarSet.empty
+        instrs
+    in
+    (* This union is needed in case a formal is not used in the instructions. *)
+    VarSet.union (all_declared_vars instrs) (VarSet.of_list formals)
+  in
+
+  (* Given a function and an array of instructions, generate a fresh version
+     for the function with these instructions. *)
+  let add_version (func : afunction) (instrs : instructions)  =
+    let label = fresh_version_label func "inlined_version" in
+    let version = {instrs = instrs; label = label; annotations = None} in
+    {func with body = version :: func.body}
+  in
+
+  (* Given `instructions` and the function identifier, extract the location,
+     return variable and arguments from the first callsite where a call is made
+     to that function *)
+  let extract_callsite instrs fun_ref : (pc * variable * argument list) =
+    let rec at pc =
+      match[@warning "-4"] instrs.(pc) with
+      | Call (x, (Simple (Constant (Fun_ref f))), es) ->
+        if f = fun_ref then (pc, x, es) else at (pc + 1)
+      | _ -> at (pc + 1)
+    in
+    at 0
+  in
+
+  (* Replace variables in `callee` instructions so that they don't match the
+     `caller` variables. The formals list of the callee is also updated
+     accordingly *)
+  let replace_vars caller ({formals; instrs} as callee) =
+    let caller_vars = function_vars caller in
+    let callee_vars = function_vars callee in
+    let replacements =
+      List.map (fun old -> (old, fresh caller_vars old))
+        (VarSet.elements callee_vars)
+    in
+    {formals = List.map (fun var -> List.assoc var replacements) formals;
+     instrs = Array.map (replace_all_vars replacements)#instruction instrs;}
+  in
+
+  (* Redirect all returns in the `callee` to a unique label (inserted later at
+     the end of the callee's body) after assigning the return expression to a
+     designated `res_var` and dropping all the variables in scope. *)
+  let replace_returns res_var ret_lab ({formals; instrs} as callee) =
+    let extract_returns instrs : (pc * expression) list =
+      let rec loop pc acc =
+        if pc = Array.length instrs then acc
+        else match[@warning "-4"] instrs.(pc) with
+          | Return e -> loop (pc + 1) ((pc, e) :: acc)
+          | _ -> loop (pc + 1) acc
+      in
+      loop 0 []
+    in
+    let scope = Scope.infer {formals = VarSet.of_list callee.formals;
+                             instrs = callee.instrs;}
+    in
+    let returns = extract_returns instrs in
+    let subst_returns (pc, e) =
+      match scope.(pc) with
+      | DeadScope ->
+        (* If it is dead scope, it remains dead scope even after this
+           substitution. This replacement is done for the sake of consistency
+           and clarity of inlined output. *)
+        (pc, 1, [|Assert (Simple (Constant (Bool false)));
+                  Assign (res_var, e);
+                  Goto ret_lab |])
+      | Scope varset ->
+        let varlist = VarSet.elements varset in
+        let drop_instrs = List.map (fun var -> Drop var) varlist in
+        let new_instrs = Array.of_list ([Assign (res_var, e)]
+                                        @ drop_instrs
+                                        @ [Goto ret_lab]) in
+        (pc, 1, new_instrs)
+    in
+    let substs = List.map subst_returns returns in
+    {callee with instrs = fst (subst_many instrs substs)}
+  in
+
+  (* Replace labels in `callee` instructions so that they don't match the
+     `caller` labels *)
+  let replace_labels caller ({formals; instrs} as callee) =
+    let used_labels = extract_labels caller.instrs in
+    let replace l = fresh used_labels l in
+    let mapper instr =
+      match[@warning "-4"] instr with
+      | Label (MergeLabel l) -> Label (MergeLabel (replace l))
+      | Label (BranchLabel l) -> Label (BranchLabel (replace l))
+      | Goto l -> Goto (replace l)
+      | Branch (e, l1, l2) -> Branch (e, replace l1, replace l2)
+      | i -> i
+    in
+    {callee with instrs = Array.map mapper instrs}
+  in
+
+  (* Inserts the header for the inlined callee body. Assigns all the formals to
+     the arguments and initializes `res_var` to `nil`*)
+  let insert_prologue res_var arguments ({formals; instrs} as callee) =
+    let vars = Array.of_list (res_var :: formals) in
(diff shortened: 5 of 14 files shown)


parser.messages
```diff
@@ -140,9 +140,9 @@ program: OSR IDENTIFIER LBRACKET NIL RBRACKET LPAREN IDENTIFIER COMMA IDENTIFIER
 
 There was an error parsing the specification of the new environment of an
 osr instruction. The specification is a comma-separated list of terms of the form
-"const x = e" (where "e" is an expression), "mut x = e", "mut x = &y"
-or just "mut x". For example,
-"osr l [(a==1)] (Main,optimized,l2) [const x = e, mut y = &y, mut z]".
+"const x = e" (where "e" is an expression), "mut x = e", "mut x = &y"
+or just "mut x". For example,
+"osr l [(a==1)] (Main,optimized,l2) [const x = e, mut y = &y, mut z]".
```
gasche: The changes look like something your editor would do automatically and they are wrong (they turn unbreakable spaces into breakable spaces). Could you revert the changes to `parser.messages`? (`git show master:parser.messages > parser.messages` would revert the file, and then you could commit that.)

aviralg: Done.

analysis.ml
```diff
@@ -180,13 +180,16 @@ end
 
 exception DuplicateFormalParameter
 
-let as_var_set (formals : formal_parameter list) =
+let as_var_list (formals : formal_parameter list) =
```
o-, line 183: I know this is not really you code. But this looks a bit fishy. I think we should:
1. Check for duplicate parameters in the `check.ml` module and not here
2. only have one of the two `as_var_list` or `as_var_set` (I don't care which one) and then when you need something different the caller just needs to `VarSet.of_list` or `VarSet.elements`

edit.ml
```diff
@@ -1,37 +1,37 @@
...
+      (fun acc instr -> VarSet.union (declared_vars instr) acc)
+      VarSet.empty instrs
+  in
+  fresh existing var
+
+let extract_labels instrs =
+  let add_label acc instr =
+    match[@warning "-4"] instr with
+    | Label (MergeLabel l) -> LabelSet.add l acc
+    | Label (BranchLabel l) -> LabelSet.add l acc
```
o-: You can do `| Label (MergeLabel l | BranchLabel l) -> ...`

instr.ml
```diff
@@ -1,12 +1,6 @@
-module Variable = struct
-  type t = string
-  let compare = String.compare
-end
+module Variable = String
```
o-, line 1: haha, thanks :)

instr.ml
```diff
@@ -306,6 +308,17 @@ let lookup_fun (prog : program) (f : identifier) : afunction =
   try List.find (fun {name} -> name = f) prog.functions with
   | Not_found -> raise (FunctionDoesNotExist f)
 
+let update_fun ({main; functions} as prog : program) (func : afunction) : program =
```
o-: maybe `replace_fun`

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
+open Instr
+open Edit
+
+(* Unlike `analysis_input`, this preserves the order of `formals`. This
+   facilitates matching the formals to the arguments passed at a callsite.*)
```
o-, line 5: I think we should probably change the type of analysis input. Obviously a set is not always optimal.

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+
+  (* Given a list of (old_name, new_name) pairs, replace all occurrences of an
+     `old_name` with the `new_name` in the instruction. Note that this process
+     has an inherent subtlety. Consider vars_alist = [(x, x_1); (x_1, x_2)].
+     All occurrences of x have to be replaced by x_1 and all occurrences of x_1
+     have to be replaced by x_2. But an occurrence of x replaced by x_1 should
+     not again be replaced by x_2. Hence the function works on the entire list
+     of replacement pairs at once rather than being called on one pair at a
+     time. *)
+  (* TODO - fix using visitor pattern if appropriate.*)
```
o-: It seems @gasche 's `Instr.map` should be used here. See `Edit.replace_uses` for an example on how to use it. The idea is that you inherit from `Instr.map` and override all the functions that correspond to the parts of instructions you want to change.

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+      let rec loop pc acc =
+        if pc = Array.length instrs then acc
+        else match[@warning "-4"] instrs.(pc) with
+          | Return e -> loop (pc + 1) ((pc, e)::acc)
+          | _ -> loop (pc + 1) acc
+      in
+      loop 0 []
+    in
+    let scope = Scope.infer {formals = VarSet.of_list callee.formals;
+                             instrs = callee.instrs;}
```
o-, line 82: I guess at this point the scope of the callee is guaranteed to be still correct in itself, because you did just renamings, right? But we got to be extra careful since it will fail on an ill-scoped function.

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+    in
+    let returns = extract_returns instrs in
+    let subst_returns (pc, e) =
+      match scope.(pc) with
+      | DeadScope ->
+        (* If it is dead scope, it remains dead scope even after this
+           substitution. This replacement is done for the sake of consistency
+           and clarity of inlined output. *)
+        (pc, 1, [|Assign (res_var, e); Goto ret_lab |])
```
o-: maybe you could emmit an `Assert (false)` instruction as well?

o- (commented):
sweet!

lgtm, except for the comments I put inline.

Don't forget to `git rebase --root -i` and squash all commits into (probably) one.

examples/inline_advanced.sou
```diff
@@ -0,0 +1,25 @@
+function main ()
+ var x = 24
+ call y1 = 'foo (x)
+ print y1
```
mhyee: For these example programs, you can use `assert`s instead of `print`s.

examples/inline_advanced.sou
```diff
@@ -0,0 +1,25 @@
+function main ()
```
mhyee, line 1: Are the inlined versions really big? If not, it would be nice to include them in the examples directory, so we can see the before/after programs.

examples/inline_advanced.sou
mhyee: I see that you're using `read_return` in the test suite. But using `assert` here allows us to test the programs with the test runner, which tries all combinations of optimizations.

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+open Edit
+
+(* Unlike `analysis_input`, this preserves the order of `formals`. This
+   facilitates matching the formals to the arguments passed at a callsite.*)
+type inlining_input = {
+  formals : variable list;
+  instrs : instructions;
+}
+
+let as_inlining_input (func:afunction) (version:version) : inlining_input =
```
mhyee: Nitpick: please put space before and after `:`.

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+(* Given a program, inline the functions in it to the maximum possible extent.
+   Transitively recursive calls are inlined until they become self recursive.
+   Self recursive callsites are not inlined. Intermediate inlined versions are
+   stored in the caller. *)
+let inline ({main; functions} as orig_prog : program) : program option =
+
+  (* Given the formals and instructions of a function, get all the declared and
+     used vars in the function. *)
+  let function_vars {formals; instrs} =
+    let all_declared_vars instrs = Array.fold_left
```
mhyee, line 24: Should this be in `instr.ml` where `all_defined_vars` and `all_dropped_vars` are defined?

Actually, are `all_defined_vars` or `all_dropped_vars` even used?

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+    (* This union is needed in case a formal is not used in the instructions. *)
+    VarSet.union (all_declared_vars instrs) (VarSet.of_list formals)
+  in
+
+  (* Given a function and an array of instructions, generate a fresh version
+     for the function with these instructions. *)
+  let add_version (func : afunction) (instrs : instructions)  =
+    let label = fresh_version_label func "inlined_version" in
+    let version = {instrs = instrs; label = label; annotations = None} in
+    {func with body = List.cons version func.body}
```
mhyee: This is a style nitpick, but we've been generally using `head :: tail` instead of `List.cons head tail`.

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+     `old_name` with the `new_name` in the instruction. Note that this process
+     has an inherent subtlety. Consider vars_alist = [(x, x_1); (x_1, x_2)].
+     All occurrences of x have to be replaced by x_1 and all occurrences of x_1
+     have to be replaced by x_2. But an occurrence of x replaced by x_1 should
+     not again be replaced by x_2. Hence the function works on the entire list
+     of replacement pairs at once rather than being called on one pair at a
+     time. *)
+  (* TODO - fix using visitor pattern if appropriate.*)
+  let replace_vars_in_instruction vars_alist instr : instruction =
+    let replace x = List.assoc x (vars_alist @ [(x, x)]) in
```
mhyee: It's more verbose, but I think it'd be clearer if you wrote (with the shorter case first)

```ocaml
let replace x =
  if not (List.mem x vars_alist)
  then x
  else List.assoc x vars_alist
```

Appending `(x,x)` feels too clever for me.

transform_inline.ml
mhyee: Yeah, looking over the function, I think all you need to override are `variable_use`, `variable_assign`, and `binder`.

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+    | Read x ->
+      Read (replace x)
+    | Label _ | Goto _ | Comment _ ->
+      assert (VarSet.is_empty (used_vars instr));
+      instr
+  in
+
+  (* Given `instructions` and the function identifier, extract information from
+     the first callsite where a call is made to that function *)
+  let extract_callsite instrs fun_ref : (pc * variable * argument list)=
```
mhyee: Nitpick: missing space before `=`.

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+      let map = List.map in_osr map in
+      Osr {label; cond; target; map}
+    | Read x ->
+      Read (replace x)
+    | Label _ | Goto _ | Comment _ ->
+      assert (VarSet.is_empty (used_vars instr));
+      instr
+  in
+
+  (* Given `instructions` and the function identifier, extract information from
```
mhyee: Please elaborate on what "information" is.

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+  in
+
+  (* Replace variables in `callee` instructions so that they don't match the
+     `caller` variables. The formals list of the callee is also updated
+     accordingly *)
+  let replace_vars caller ({formals; instrs} as callee) =
+    let caller_vars = function_vars caller in
+    let callee_vars = function_vars callee in
+    let replacements =
+      List.map (fun old -> (old, Edit.fresh caller_vars old))
```
mhyee: `fresh` instead of `Edit.fresh`.

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+        (VarSet.elements callee_vars)
+    in
+    {formals = List.map (fun var -> fresh caller_vars var) formals;
```
mhyee: It makes me a bit nervous that `fresh` is being called a second time here, and that the renamed formals will match up with the renamed function variables. If `fresh` was implemented as something like `gensym` that kept a global counter, then this would break.

`replacements` should include renamings for the formals, so maybe you could use that? Again, it would be more verbose, but then you wouldn't need to call `fresh` twice.

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+     instrs = Array.map (replace_vars_in_instruction replacements) instrs;}
+  in
+
+  (* Redirect all returns in the `callee` to a unique label (inserted later at
+     the end of the callee's body) after assigning the return expression to a
+     designated `res_var` and dropping all the variables in scope. *)
+  let replace_returns res_var ret_lab ({formals; instrs} as callee) =
+    let extract_returns instrs : (pc * expression) list=
```
mhyee: Nitpick: missing space before `=`.

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+    let callee_vars = function_vars callee in
+    let replacements =
+      List.map (fun old -> (old, Edit.fresh caller_vars old))
+        (VarSet.elements callee_vars)
+    in
+    {formals = List.map (fun var -> fresh caller_vars var) formals;
+     instrs = Array.map (replace_vars_in_instruction replacements) instrs;}
+  in
+
+  (* Redirect all returns in the `callee` to a unique label (inserted later at
```
mhyee: Mention `ret_lab` in the comment?

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+    let substs = List.map subst_returns returns in
+    {callee with instrs = fst (Edit.subst_many instrs substs)}
+  in
+
+  (* Replace labels in `callee` instructions so that they don't match the
+     `caller` labels *)
+  let replace_labels caller ({formals; instrs} as callee) =
+    let used_labels = Edit.extract_labels caller.instrs in
+    let replace l = fresh used_labels l in
+    let mapper instr =
```
mhyee: You could probably use `Instr.map` here and override `goto_label` and `branch_label`.

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+                       Drop res_var|]
+    in
+    {callee with
+     instrs = fst (Edit.subst instrs (Array.length instrs) 0 new_instrs)}
+  in
+
+  (* Transforms the callee instructions to a form that can be substituted
+     inside the caller.*)
+  let compose caller callee ret_var arguments : inlining_input =
+    let callee' = callee
```
mhyee: I think it's fine to shadow, i.e. use `let callee = callee |> replace_vars caller ...` since you never use the original `callee` again.

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+    in
+    (* It is important to generate return label and result variable after
+       replacing variables and labels. Otherwise there is a good chance that
+       they will not remain fresh. For ex - if `res` is present in both caller
+       and callee, then `res_1` as a result variable will not remain fresh once
+       `replace_vars` is called. Because `replace_vars` will replace `res` in
+       callee with `res_1` and `res_1` is already used for the result variable.
+       This problem will not happen if we `replace_vars` before generating a
+       fresh `res_var`. *)
+    let ret_lab = "inl"
```
mhyee: Couldn't you combine the callee and caller instructions, so that `fresh_label` is only called once?

Similarly for `res_var` below.

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+      | Goto l -> Goto (replace l)
+      | Branch (e, l1, l2) -> Branch (e, replace l1, replace l2)
+      | i -> i
+    in
+    {callee with instrs = Array.map mapper instrs}
+  in
+
+  (* Inserts the header for the inlined callee body. Assigns all the formals to
+     the arguments and initializes `res_var` to `nil`*)
+  let insert_prologue res_var arguments ({formals; instrs} as callee)=
```
mhyee: Nitpick: missing space before `=`.

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+    in
+    callee'
+    |> replace_returns res_var ret_lab
+    |> insert_prologue res_var arguments
+    |> insert_epilogue res_var ret_var ret_lab
+  in
+
+  (* Given the caller and callee identifiers, generate a fresh version of caller
+     with the callee inlined.*)
+  let inline_pair (clr : identifier) (cle : identifier) (prog : program) =
```
mhyee: Can you think of better names? It's a bit messy to have, e.g. `cle`, `callee`, `callee'`, and `callee''`. (Personally, I find even `clr` and `cle` to be too terse.)

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+    List.map (fun (clr, cle, _ ) -> (clr, cle) ) (List.sort comp ord)
+  in
+
+  (* Given a list of caller - callee pairs, inline the callee inside the caller
+     until the list is exhausted. Order matters, as only this order will result
+     in a maximally inlined program excluding recursive calls. The final program
+     will only have recursive calls left to be inlined. *)
+  let inline_with order prog =
+    List.fold_left
+      (fun p (clr, cle) -> update_fun p (inline_pair clr cle p))
```
mhyee: I'd prefer `caller` and `callee`. `clr` and `cle` don't save all that much.

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+    let rec inspect_instr ((ord, clr, vis, red, dep) as acc) instr =
+      match[@warning "-4"] instr with
+      | Call (x, (Simple (Constant (Fun_ref f))), es) ->
+        (* If `f` is already visited in this branch, then there is recursion,
+           so don't add this pair to accumulator.*)
+        if (VarSet.mem f vis) then acc
+        (* If `f` is already analyzed for inlining, then add it but don't go
+           inside it.*)
+        else if (VarSet.mem f red)
+        then (List.cons (clr.name, f, dep) ord, clr, vis, red, dep)
```
mhyee: `(clr.name, f, dep) :: ord`

transform_inline.ml
```diff
@@ -0,0 +1,312 @@
...
+   list of these pairs such that the inlining happens in a bottom up fashion.
+   If a caller-callee pair is already encountered, then it is not included
+   again. This is because once the callee has been inlined in the caller at a
+   given callsite, then the callsite no longer exists. However, multiple
+   callsites with the same target will result in as many caller-callee pairs.
+   Note - The word reduced is used in the general sense of something being
+   simplified. If a function has been analyzed for inlining, its considered
+   reduced. In reality, inlining expands this function. *)
+  let compute_inlining_order (prog : program) : (identifier * identifier) list =
+    let rec inspect_instr ((ord, clr, vis, red, dep) as acc) instr =
```
mhyee: I'm having a hard time following this function. (It doesn't help that it's late and this is the last function I have to review.)

In particular, I can't follow the terse names like `ord`, `clr`, `vis`, `red`, and `dep`. (`acc` is commonly used so it's fine.)

Also, passing a 5-tuple around is kind of awkward; I'd be tempted to just define a new record type. That would help with readability, and also in the places where you only need to change one field.

Finally, I think it's a bit excessive when you have names like `ord`, `ord'`, `ord''`, and `ord'''`.

mhyee (changes requested):
This is really cool! And thanks for the detailed comments!

I'd like to see an example where constant folding (or some other optimization) can be applied after inlining. Right now, none of the inlining examples allow for multiple optimizations.

I'd like to try running inlining on the vector example, but it needs to be rebased/updated, so I think I'll wait until this PR is merged.

I'm requesting changes because I'd like to look at `compute_inlining_order` again, when I'm less tired and its names are better.

analysis.ml
aviralg: `as_var_list` has the advantage of maintaining the order of `formals` which is lost as soon as conversion to `Set` is performed. If we intend to change the type of `analysis_input` to contain a `List` of `formals` rather than a `Set` of `formals`, then `as_var_list` makes more sense.

instr.ml
aviralg: I did that because there is a function called `lookup_fun` and `update` seems to go along well with `lookup`. But its really a minor thing. Changed it to `replace_fun`.

transform_inline.ml
aviralg: Good catch, I intended to replace that, but forgot. Corrected it now.

transform_inline.ml
aviralg: This one looks a bit long and requires 2 lookups. A better version would be the following -
```
try List.assoc x vars_alist with
| _ -> x
```
I had this version initially but felt that the clever approach was readable.

examples/inline_advanced.sou
aviralg: They are big.

instr.ml
aviralg: Welcome :-) . Gabriel suggested this while I was unifying the freshen family of functions. He said that - "its dirty but its you who is doing it". I guess the joke's on him now ;-)

transform_inline.ml
aviralg: I can do that in a separate PR.

transform_inline.ml
aviralg: Removed them. They are not used.

aviralg: This will be moved to `instr.ml` once we change `analysis_input` to contain a `List` of `formals`.

transform_inline.ml
aviralg: Yup. That's why `replace_vars` also replaces `formals`' `names`.

aviralg (commented):
Thanks for the reviews guys. I believe I have resolved most of the issues.

transform_inline.ml
aviralg: The idea was to use `clr` and `cle` for just the identifiers and not the actual objects.

transform_inline.ml
aviralg: Done.

gasche:
> A better version would be the following -
>
>     try List.assoc x vars_alist with
>     | _ -> x

Catching all exceptions is not very good style, it's better to be specific in the exception name

    try List.assoc x vars_alit with Not_found -> x

gasche:
In the interest of keeping the main repository somewhat bisectable, I would be interesting in having the series of commits reworked into a few well-separated commits such that each commit leaves the repository in a correct (buildable, testable) state. If you wish, we can do this together on Monday morning. I feel that the PR is in an essentially mergeable state now, but I like this to happen before merging.

mhyee:
Thanks for the updates. I still have a few remaining comments:

1. Is the `inline_basic` example small enough that you could add the before/after?

2. I'd like to see an inlining example that allows constant folding (or some other optimization).

3. The extra documentation for `compute_inlining_order` helps, but I'm still going to push for using a record type and using better variable names. (As you've probably noticed, I'm not a big fan of terse names, and I think a name like `foobar''` (with two primes) is already going too far.)

> In the interest of keeping the main repository somewhat bisectable, I would be interesting in having the series of commits reworked into a few well-separated commits such that each commit leaves the repository in a correct (buildable, testable) state. If you wish, we can do this together on Monday morning. I feel that the PR is in an essentially mergeable state now, but I like this to happen before merging.

Would squashing everything into a single commit be OK, or do you prefer more granular commits? I'd be fine with a single commit -- it would be a large commit, but I don't see any obvious place to break it up.

gasche:
I'm ok with a big squash, but we could also split (eg. the `fresh` refactoring can go separately).

analysis.ml
o-: agreed

o-:
Squash, merge, fix the rest later, and repeat :)

aviralg:
@mhyee - Regarding the `compute_inlining_order` function -
1) The initial version had full names rather than mnemonics, but that was causing quite a few splits in the lines to respect the 80 character limit. I chose these mnemonics because they are obvious and make the code compact enough without exceeding the screen size.
2) I don't see a compelling reason to use a record here. This is a perfect case for a `tuple`. I don't want to create a record type just for one function. Besides it would be used no differently than a `tuple`. 
3) The terse names are mnemonics and the documentation gives the full forms. `vis` for visited and `red` for reduced are readable.
4) I renamed some variables, I believe most of your complaints will be gone now.

aviralg:
@gasche  @mhyee  @o-  I would squash all in a single commit just because its quick, easy and also because the change to `fresh` family of functions is minor refactoring. But I am also open to having 2 commits, one that refactors `fresh` and friends, and the other that implements inlining, though that might be more difficult for me due to my poor `git-fu` (I would be happy to know the ways of the masters).

mhyee:
I still disagree about `compute_inlining_order`, but in the interest of moving forward, I'll approve the changes.

mhyee (approved):

gasche:
👏 

(Rebase didn't work, we had to `export HASH=$(git rev-parse HEAD); git reset master; git commit -a -C $HASH`.)

o-:
ah probably because of the merge commits in there...

PR #148, opened 2017-06-05
don't wrap exceptions

o-:
the wrapping of the exception makes it harder to trace back to
the point of origin of the exception. Since all those exceptions
in the optimizer should be fixed it does not make much sense
to have it there. When I want to debug a problem in an optimizer
I always have to remove this code first to be able to see the
origin of the exn.

### DIFF
--- a/transform.ml
+++ b/transform.ml
@@ -44,13 +44,7 @@ let as_opt_program (transform : opt_function) : opt_prog =
     | Some instrs -> true, instrs in
   fun prog ->
     let changed, main =
-      try transform' prog.main
-      with exn ->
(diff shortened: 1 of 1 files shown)


mhyee (approved):
I'm surprised that this is a problem (or maybe I don't understand exceptions), because the exception is being re-raised.

In any case, I think this is a reasonable change.

gasche:
(The not-yet-released OCaml 4.05 has a feature to be very precise about backtrace re-raising that would make it a non-issue, but let's merge this for now -- we can revisit it after the release :- )

PR #149, opened 2017-06-05
inlining: fix var and label renaming

o-:
To rename variables and labels we cannot call `fresh` on all
existing variables in parallel. For example consider the function:

```
caller:
  var x = 1

callee:
  var x   = 1
  var x_1 = 2
```

In general any try to do the renaming in parallel makes unreasonable
assumptions about how `fresh` is implemented. For example it would be
perfectly reasonable for `fresh` to always return `foo` if it's fresh.

Thus the only way of using fresh which does not make assumptions on
how it works is to always feed back the generated names to avoid generating
the same name again.

### DIFF
--- a/edit.ml
+++ b/edit.ml
@@ -9,7 +9,16 @@ let fresh (existing : StringSet.t) (name : string) : string =
   let rec find i =
     let cand_var = cand i in
     if is_fresh cand_var then cand_var else find (i+1) in
-  find 1
+  if is_fresh name
+  then name
+  else find 1
+
+let fresh_many existing names =
+  let replace (used, acc) old =
+    let fresh_var = fresh used old in
+    (VarSet.add fresh_var used, (old, fresh_var) :: acc)
+  in
+  snd (List.fold_left replace (existing, []) names)
 
 let fresh_var instrs var =
   let existing = Array.fold_left
--- a/transform_inline.ml
+++ b/transform_inline.ml
@@ -1,5 +1,6 @@
 open Instr
 open Edit
+open Types
 
 (* Unlike `analysis_input`, this preserves the order of `formals`. This
    facilitates matching the formals to the arguments passed at a callsite.*)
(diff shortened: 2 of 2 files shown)


transform_inline.ml
```diff
@@ -107,8 +111,16 @@ let inline ({main; functions} as orig_prog : program) : program option =
      `caller` labels *)
   let replace_labels caller ({formals; instrs} as callee) =
     let used_labels = extract_labels caller.instrs in
-    let replace l = fresh used_labels l in
+    let callee_labels = LabelSet.elements (extract_labels callee.instrs) in
+    let replacements =
```
aviralg: Given that this pattern is repeating itself both on line 60 and 116, we can factor it out as a function called `fresh_many` (just like `subst` is generalized to `subst_many`) in `edit.ml`.

edit.ml
```diff
@@ -6,6 +6,7 @@ module StringSet = Set.Make(String)
 let fresh (existing : StringSet.t) (name : string) : string =
   let cand i = name ^ "_" ^ (string_of_int i) in
   let is_fresh cand_var = not (StringSet.mem cand_var existing) in
+  if is_fresh name then name else
```
aviralg: Thanks, this is useful.

aviralg (approved):
Looks good. You are right, this code can't make assumptions on how `fresh` works.

edit.ml
gasche: I would find it easier to read if the `if` was on the `find 1` call.

transform_inline.ml
```diff
@@ -56,10 +57,13 @@ let inline ({main; functions} as orig_prog : program) : program option =
      accordingly *)
   let replace_vars caller ({formals; instrs} as callee) =
     let caller_vars = function_vars caller in
-    let callee_vars = function_vars callee in
+    let callee_vars = VarSet.elements (function_vars callee) in
     let replacements =
-      List.map (fun old -> (old, fresh caller_vars old))
-        (VarSet.elements callee_vars)
+      let replace (used , acc) old =
```
mhyee: Nitpick: extra space before `,`.

transform_inline.ml
```diff
@@ -56,10 +57,13 @@ let inline ({main; functions} as orig_prog : program) : program option =
...
+        let fresh = fresh used old in
```
mhyee: I think it would be clearer to name it `fresh_var`. (Yeah, I don't like shadowing/reusing names.)

mhyee (approved):

gasche:
The resulting patch looks nice, thanks!

PR #150, opened 2017-06-06
Remove broken check.

mhyee:
This check is performed a few lines below. Worse, the check is missing
begin/end, so nothing actually happens to the input programs.


o-:
ehm, this was there intentionally (unfinished though)...

There is a chicken-egg problem.

1. autofix breaks on broken programs
2. check complains about broken scope

so the idea was to run check first but only report errors that autofix would crash on, then autofix, then re-run checks and report all errors.

mhyee:
Opened a new issue for tracking: #152.

I had to remove it because it was blocking me. I was very confused when I ran sourir and nothing happened.

PR #151, opened 2017-06-06
fix constantfolding type confusion

o-:
constant fold would happily convert `a[b]` into `nil[b]`. The main
reason is that our operations where not "typed" and therefore we
had no way of specifying which position should be a variable or an
arbitrary expression.

This commit changes the structure of expressions to syntactically
enforce the correct argument types.

### DIFF
--- a/check.ml
+++ b/check.ml
@@ -86,9 +86,9 @@ let well_formed prog =
         | Unop (_op, e) ->
           check_simple_expr e
         | Binop (_op, e1, e2) ->
-            check_simple_expr e1;
-            check_simple_expr e2
-        | Array_index(_, i) ->
+          check_simple_expr e1;
+          check_simple_expr e2
+        | Array_index (_, i) ->
           check_simple_expr i
         | Array_length (e) ->
           check_simple_expr e in
--- a/check.ml
+++ b/check.ml
@@ -83,8 +83,15 @@ let well_formed prog =
       in
       let check_expr = function
         | Simple e -> check_simple_expr e
-        | Op (_op, xs) ->
-          List.iter check_simple_expr xs in
+        | Unop (_op, e) ->
+          check_simple_expr e
+        | Binop (_op, e1, e2) ->
+            check_simple_expr e1;
+            check_simple_expr e2
+        | Array_index(_, i) ->
+          check_simple_expr i
+        | Array_length (e) ->
+          check_simple_expr e in
       let check_arg = check_expr in
       let check_osr = function
         | Osr_var (_, e) -> check_expr e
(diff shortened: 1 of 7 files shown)


gasche:
The other option is to just extend `x[e]` to `e1[e2]`, are you sure we want to go towards the more specific?

o-:
> The other option is to just extend x[e] to e1[e2]

yes, and I would even welcome that change.

> are you sure we want to go towards the more specific?

At least to me the code seems a lot cleaner this way.

gasche:
I liked the idea of having a generic template that could be added to with very little changes to the codebase (except the parsing and printing that remain operation-specific). Using GADTs I could actually make the generic template typed, but I wanted to keep the type-system sophistication as low as possible for accessibility reasons.

Would you like me to generalize array access and implement the GADT solution to keep a single `Op` constructor but have type-level tracking of the arity?

mhyee:
> The other option is to just extend x[e] to e1[e2],

Wouldn't that allow expressions like `x[1][2]`? I don't really like that.

> Would you like me to generalize array access and implement the GADT solution to keep a single Op constructor but have type-level tracking of the arity?

If it's not too much work, sure. At this point, given two correct solutions, I would choose the lower effort one. The advantage of this PR is that it's already implemented. (But I haven't reviewed it yet.)

gasche:
> Wouldn't that allow expressions like `x[1][2]`? I don't really like that.

Not really, the simplest extension is to expect a `simple_expression` on the left.

o-:
I really don't have an opinion on it. All of the proposed solutions are fine. As long as the bug gets fixed :)

check.ml
```diff
@@ -83,8 +83,15 @@ let well_formed prog =
       in
       let check_expr = function
         | Simple e -> check_simple_expr e
-        | Op (_op, xs) ->
-          List.iter check_simple_expr xs in
+        | Unop (_op, e) ->
+          check_simple_expr e
+        | Binop (_op, e1, e2) ->
+            check_simple_expr e1;
```
mhyee: Nitpick: alignment

check.ml
```diff
@@ -83,8 +83,15 @@ let well_formed prog =
...
+            check_simple_expr e2
+        | Array_index(_, i) ->
```
mhyee: Nitpick: missing space

mhyee (approved):
Since this is a reasonable fix to the bug and is implemented, I think we should just go with it.

If you want, I can fix the nitpick issues and then merge it myself.

gasche:
Ok, let's merge.

PR #153, opened 2017-06-06
use the OCaml native compiler

gasche:
The bytecode compiler compiles faster (and is more incremental) than
the native compiler, so I guess I used it at the start for convenience
of development. Given that build times remain small and that we are
starting to hit performance issues, let's move to the real deal now --
as we probably should have a while ago.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,12 +16,12 @@ OCAMLBUILD+= -plugin-tag 'package(bisect_ppx.ocamlbuild)'
 all: tests test_examples sourir
 
 tests:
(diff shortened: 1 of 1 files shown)


mhyee (approved):
Vector examples are about 3x faster now.

Full build with native compiler:
```
real	0m8.723s
user	0m3.005s
sys	0m1.303s
```

Full build with bytecode compiler:
```
real	0m4.120s
user	0m1.713s
sys	0m0.767s
```

o-:
nice

PR #160, opened 2017-06-06
Inline osr

o-:
Not so well tested yet....

...but still worth a look.


transform_inline.ml
```diff
@@ -103,8 +86,7 @@ let inline ({main; functions} as orig_prog : program) : program option =
 
   (* Replace labels in `callee` instructions so that they don't match the
      `caller` labels *)
-  let replace_labels caller ({formals; instrs} as callee) =
-    let used_labels = extract_labels caller.instrs in
+  let replace_labels used_labels ({formals; instrs} as callee) =
     let callee_labels = LabelSet.elements (extract_labels callee.instrs) in
     let replacements = Edit.fresh_many used_labels callee_labels in
```
aviralg: I think you can remove the `Edit` as it is already opened.

o-:
@aviralg to reiterate what gabriel said by mail: thanks a lot for the great work on the inliner. I hope you don't mind that I had to replace your inline order logic. The problem was that it was really hard to keep track of extra information while discovering the callgraph. That's why I chose to replace the inline order list by a more graph like structure that I can just traverse. The actual inline logic is still there and proved to be super robust.

aviralg (commented):
Its great! I found a couple of formatting nitpicks but I would leave them in the able hands of @mhyee .

o-:
There might be some rough edges, but it should be ready as a basis to talk about it...

The main idea is that our osr gets extended to the right by an optional list of additional continuations to create. for example `(a,b,c,x)[var y = 1]` creates the continuation `(a,b,c), x, (y->1)`. 

The most interesting part is [here](https://github.com/reactorlabs/sourir/pull/160/commits/54ff835d023d5c0341dd30e95fa6f8a8dd34af2e#diff-ddb3aa6ec7a2afe48b9b7aee59c17b26R181) where the current activation frame osr mapping is converted into an continuation creating entry that will be appended to the inlinee osr's.

aviralg:
@o- don't worry, I don't mind. Code has to be updated as the requirement changes.

instr.ml
```diff
@@ -413,7 +420,10 @@ class map = object (m)
       Length (m#expression e)
     | List es ->
       List (List.map m#expression es)
-  method osr_map = function
+
+  method osr_frame ({var_map} as frame) = { frame with var_map = m#osr_var_map var_map; }
+  method osr_var_map = List.map m#osr_var_map_entry
+  method osr_var_map_entry = function
```
mhyee: For consistency, I think this should be called `osr_def`.

instr.ml
```diff
@@ -398,12 +404,13 @@ class map = object (m)
       Assert (m#expression e)
     | Stop e ->
       Stop (m#expression e)
-    | Osr {label; cond; target; map} ->
+    | Osr {label; cond; target; var_map; frame_maps} ->
       Osr {
         label;
```
mhyee: Sorry, but why do we *never* need a method for `label`? Likewise for `target`.

instr.ml
```diff
@@ -413,7 +420,10 @@ class map = object (m)
       Length (m#expression e)
     | List es ->
       List (List.map m#expression es)
-  method osr_map = function
+
+  method osr_frame ({var_map} as frame) = { frame with var_map = m#osr_var_map var_map; }
```
mhyee: Same question, why do we never need a method for `continuation_pos` or `continuation_res`?

tests.ml
```diff
@@ -1886,7 +1886,9 @@ let suite =
    "parser3">:: test_parse_disasm  ("var x = (y + x)\n");
    "parser4">:: test_parse_disasm  ("x <- (x == y)\n");
    "parser5">:: test_parse_disasm  ("# asdfasdf\n");
-   "parser5b">:: test_parse_disasm ("osr l [(x == y)] (f, v, l) [var x = x, var v, var x = (1+2)]\nl:\n");
+   "parser5b">:: test_parse_disasm ("osr l [(x == y)] (f, v, l) [var x = x, var v, var x = (1+2)]\n");
+   "parser5c">:: test_parse_disasm ("osr l [(x == y)] (f, v, l) [var x = x, var v, var x = (1+2)] (f,v,l,x) []\n");
+   "parser5d">:: test_parse_disasm ("osr l [(x == y)] (f, v, l) [var x = x, var v, var x = (1+2)] (f,v,l,x) [], (f,v,l,x) [var c = 4]\n");
```
mhyee: I'll just put this comment here, but there's a mistake in the commit message. I was trying to use it to figure out the syntax of the new `osr` instruction. It was missing the list of conditions:

`osr l [...] (...) [...]  (f,v,l,x)[var y = 2], (...)[...]`

eval.ml
```diff
@@ -372,25 +372,40 @@ let reduce conf =
...
+  | Osr {cond; target={func;version; pos=label}; var_map; frame_maps} ->
     let triggered = List.exists (fun cond -> get_bool (eval conf cond)) cond in
     if not triggered then
       { conf with
         pc = pc';
       }
     else begin
-      let osr_env, heap' = build_osr_frame map conf.env conf.heap in
+      let osr_env, heap = build_osr_frame var_map conf.env conf.heap in
+      let create_extra_frames (heap, frames) {continuation_pos; continuation_res; var_map} =
```
mhyee: For consistency, I suppose this should be called `build_extra_frames`.

Also, can you move the definition up, to where `build_call_frame` and `build_osr_frame` are defined? That will keep this section a little cleaner.

mhyee: I guess the `List.fold_left ...` part should also be moved out, so we just have a single line

```ocaml
let heap, extra_frames = build_extra_frames frame_maps in
```

(Plus whatever extra args it needs.)

eval.ml
```diff
@@ -372,25 +372,40 @@ let reduce conf =
...
+      let heap, extra_frames =
+        List.fold_left create_extra_frames (heap, []) frame_maps in
       let func = Instr.lookup_fun conf.program func in
       let version = Instr.get_version func version in
       let instrs = version.instrs in
       { conf with
         pc = resolve_osr instrs label;
         env = osr_env;
-        heap = heap';
+        heap;
```
mhyee, line 406: I'd prefer this to be more explicit (like `instrs` below): `heap = heap;`

transform_assumption.ml
```diff
@@ -210,23 +210,23 @@ let hoist_assumption : transform_instructions = fun ({instrs; _} as inp) ->
...
         let cond_vars = expr_vars c in
         begin match find_candidate_osr c cond_vars (pc-1) None with
         | None -> true
         | Some pc' ->
           changed := true;
           begin match[@warning "-4"] instrs.(pc') with
-          | Osr {label; cond; target; map} ->
-            instrs.(pc') <- Osr {label; cond = c::cond; target; map}
+          | Osr {label; cond; target; var_map; frame_maps} ->
+              instrs.(pc') <- Osr {label; cond = c::cond; target; var_map; frame_maps}
```
mhyee: Nitpick: alignment

transform_assumption.ml
```diff
@@ -32,11 +32,11 @@ let insert_checkpoints (func:afunction) =
           version=version.label;
           pos=checkpoint_label pc;
         } in
-        InsertBefore [Osr {label=checkpoint_label pc; cond=[]; target; map=osr};]
+        InsertBefore [Osr {label=checkpoint_label pc; cond=[]; target; var_map=osr; frame_maps=[]};]
     in
-    if pc = 0 then Unchanged else
```
mhyee: Why was this removed?

transform_assumption.ml
```diff
@@ -32,11 +32,11 @@ let insert_checkpoints (func:afunction) =
...
     match[@warning "-4"] instrs.(pc) with
-    | Stop _ | Return _ | Label _ | Comment _ -> Unchanged
+    | Label _ | Comment _ ->
```
mhyee, line 39: Why was this changed?

transform_inline.ml
```diff
@@ -154,108 +144,103 @@ let inline ({main; functions} as orig_prog : program) : program option =
...
+    else
+      let used_labels = ref (extract_labels cur.instrs) in
+      let get {target; next; pos; ret; args; osr} =
+        let apply next = if next = []
+                         then as_inlining_input target
+                         else apply_inlinings target next in
+        let callee = apply next in
+        match needs_osr callee.instrs, osr with
+        | false, _ ->
+            let inlinee, new_used = compose cur callee !used_labels ret args in
```
mhyee: Nitpick: alignment

transform_inline.ml
```diff
@@ -154,108 +144,103 @@ let inline ({main; functions} as orig_prog : program) : program option =
...
-  let inline_with order prog =
-    List.fold_left
-      (fun p (caller, callee) -> replace_fun p (inline_pair caller callee p))
-      prog
-      order
+  let inline_at func =
+    let inline_order = compute_inline_order func LabelSet.empty in
+    (* If there are no caller-callee pairs to inline, return `None`, else return
+       the completely inlined program *)
+    if inline_order = [] then None
```
mhyee:
```ocaml
if inline_order = []
then None
else
  ...
```

transform_inline.ml
```diff
@@ -22,6 +12,9 @@ type inlining_candidate = {
 }
 and inlining_site = inlining_candidate list
 
+let as_inlining_input (func : afunction) =
```
mhyee: This is basically `Analysis.as_analysis_input` except that it uses the active version, rather than a given version.

mhyee: I would probably just remove this and use `Analysis.as_analysis_input`.

transform_inline.ml
```diff
@@ -162,18 +155,31 @@ let inline ({main; functions} as orig_prog : program) : program option =
...
+    let version = (active_version func) in
+    let instrs = version.instrs in
     let inlinings = ref [] in
+    let create_osr_continuation top_frame call_var osr_label =
+      let pos = {
+        func = func.name;
+        version = version.label;
+        pos = osr_label; } in
+      { var_map = top_frame;
+        continuation_res=call_var;
```
mhyee: Nitpick: missing space before/after `=`. Also on the line below.

transform_inline.ml
```diff
@@ -194,6 +200,19 @@ let inline ({main; functions} as orig_prog : program) : program option =
     Array.exists is_osr
   in
 
+  let fixup_osr extra_frames input =
```
mhyee: Can you add a short comment documenting this?

transform_inline.ml
```diff
@@ -211,9 +230,9 @@ let inline ({main; functions} as orig_prog : program) : program option =
             used_labels := new_used;
             (pos, 1, inlinee.instrs)
         | true, Some osr ->
-            (* TODO(osr): Here we need to update the inlinee osr points to create the
-             *            additional osr frame. *)
-            (pos, 0, [| |])
+            let inlinee, new_used = compose cur callee !used_labels ret args in
```
mhyee: Nitpick: alignment

transform_inline.ml
```diff
@@ -154,108 +137,129 @@ let inline ({main; functions} as orig_prog : program) : program option =
...
+        match needs_osr callee.instrs, osr with
+        | false, _ ->
+            let inlinee, new_used = compose cur callee !used_labels ret args in
+            used_labels := new_used;
+            (pos, 1, inlinee.instrs)
+        | true, Some osr ->
+            let inlinee, new_used = compose cur callee !used_labels ret args in
+            used_labels := new_used;
+            (pos, 1, fixup_osr osr inlinee)
+        | true , None ->
```
mhyee: Nitpick: extra space before `,`

mhyee (commented):
This looks pretty good! I'll have to look at this again, when I'm less tired and have also looked at the spec in #129 to refresh my memory.
Tests would also be nice.

Also, as a general comment, the semantics doc has fallen behind. We should update it soon. (But it's good that the inlining spec is somewhere, even if it's a PR #129.)

instr.ml
o-: That's not a question I would deem valid. It's easy to add. My guideline is: Why do we need it *now*. My software engineering practice and also training has thought me to never ever add features just because they might come handy in the future. Like in this case I had to rewrite the existing code that was too flexible for it's own good. If I spent more than 2' trying to figure out how to abstract it even though I am sure we won't use it in the next 3 weeks that would be a waste. But what's worse is I add a badly tested (since unused) feature to the codebase. The more (untested) things in the codebase the harder it is to move it. I know that is now a completely overblown answer to this small point. But I thought it's a good situation to explain myself and why I sometimes remove things that I see as cruft.

instr.ml
o-: see above. Even if I wanted I don't see a usecase for rewriting them.

instr.ml
o-: maybe everything should be called map entry?

tests.ml
o-: ups I'll try to fix it

eval.ml
o-: sounds good

eval.ml
o-: that's kind of a very stylistic request, not?

transform_assumption.ml
o-: Why was it there in the first place :) I don't know.

Maybe it was because it could go wrong if the first instruction was a label. But since we forbid this now I don't see a reason for this case.

transform_assumption.ml
o-: To have more safepoints available. I noticed that in the case of:
```
call x = foo
return
```
there was no safepoint after the call. but we need one. Not sure if there was ever a good reason for the list.

transform_assumption.ml
o-: thanks, somehow merlin update changed identation after `->` from 2 to 4 spaces :/

o-:
Ok I reworked it quite a bit after all your suggestions.

Also I made one final change to the osr extra frames. In the spirit of my original proposal the syntax is now:

```
osr sp1 [cond] (target) [top_frame]  (cont1) [var ret1 = $, frame1], (cont2) [var ret2 = $, frame2]
```

where 
* `sp1` is the label
* `cond` is a list of trigger conditions
* `target` is the target tripple where control resumes
* `top_frame` is a variable mapping for the top most frame

the rest is a list of additional frames to create where
* `contx` is a target tripple of the return address
* `retx` is the name of the variable that receives the result of the call (ie. $ is the 'return register')
* `framex` is the variable mapping to create the additional frame

tests.ml
mhyee: The updated commit message is even better than before. Thanks!

eval.ml
mhyee: Sure, it's not a big deal either way.

transform_assumption.ml
mhyee: OK, I forgot that we don't allow the first instruction to be a label.

transform_inline.ml
```diff
@@ -187,30 +199,63 @@ let inline ({main; functions} as orig_prog : program) : program option =
...
+   *  version v1:
+   *    ...
+   *    call res = foo _
+   *    osr t0 _ (f0,v0,t0) [frame0]  (f1,v1,t1) [frame1], (f2,v2,t2) [frame2]
+   *
+   * Then we want to extend foo's extra_frames list by the following list:
+   *    (f0,v1,t0) [var res = $, frame0/res], (f1,v1,t1) [frame1], (f2,v2,t2) [frame2]
+   *
+   * This ensures that when the deoptimized foo returns it returns after the call with the call
+   * stack identical to before the call and the result stored to y.
```
mhyee: `y` -> `res` ?

examples/inline_osr.sou
```diff
@@ -0,0 +1,15 @@
+function main ()
```
mhyee, line 1: It would be nice to include an "after" version so that we don't have to run the program. Unless it's too large?

mhyee: Or at the very least, a comment explaining which opts to run (since we won't be able to dig out the commit message).

transform.ml
```diff
@@ -119,8 +121,12 @@ let optimize (opts : string list) (prog : program) : program option =
       as_opt_program const_fold
     | "prune" ->
       as_opt_program branch_prune
-    | "inline" ->
-      Transform_inline.inline
+    | "inline_max" ->
```
mhyee, line 124: There are some typos in the commit message (or code):

`inlining_*` -> `inline_*`
`inlining_map` -> `inline_max`

Is `inline_med` to a depth of 3 or 4?

transform.ml
```diff
@@ -119,8 +121,12 @@ let optimize (opts : string list) (prog : program) : program option =
...
+      Transform_inline.inline ()
+    | "inline_med" ->
+      Transform_inline.inline ~max_depth:4 ~max_size:220 ()
```
mhyee, line 127: Just curious, but where did the values 170 and 220 come from?

transform_assumption.ml
```diff
@@ -47,17 +47,64 @@ let insert_checkpoints (func:afunction) =
...
-  let transform pc =
-    match[@warning "-4"] input.instrs.(pc) with
-    | Osr {cond} when cond = [] -> Remove 1
-    | _ -> Unchanged
+module TargetSet = Set.Make(LabelTarget)
+
+(* Removes all empty checkpoints with no incoming references *)
+let remove_empty_osr : opt_prog = fun prog ->
+  (* Collect the targets (including the extra frames continuations) of all non empty osrs *)
+  let used =
```
mhyee: Could you use different names for the different variations of `used`? I would find it easier to follow.

mhyee: At least the definition of `add` is directly above its use. That I can live with.

transform_assumption.ml
```diff
@@ -47,17 +47,64 @@ let insert_checkpoints (func:afunction) =
...
+      let add tgs (v : version) = TargetSet.union tgs (used v.instrs) in
+      List.fold_left add TargetSet.empty func.body
+    in
+    let add tgs (f : afunction) = TargetSet.union tgs (used f) in
+    List.fold_left add (used prog.main) prog.functions
+  in
+
+  (* Remove all empty and unused osr instructions in all versions *)
+  let changed = ref false in
+  let transform prog =
```
mhyee: Same with `transform`.

mhyee (commented):
I like the new syntax for frames. `(f,v,t,res)` always seemed too similar to `(f,v,t)` for me.

Do you know how long the vector example took under the small and medium inlining? (If you don't have the numbers, I'll just run it myself later.)

o-:
> Do you know how long the vector example took under the small and medium inlining? (If you don't have the numbers, I'll just run it myself later.)

about 4-5' for small and a bit more than a minute for the long test runner

Update: ups sorry. those numbers are for the whole testsuite including vector

examples/inline_osr.sou
o-: not sure if I understand the concerns? This is run by the test runner, that was the main concern. Any optimizations combined with inline (obvious from the file name) is interesting. Why an after? I can just run it...

transform.ml
o-: 170 - 220 = 390 which is 10 from 400 which is a round number

o-: (ie. out of thin air)

transform_assumption.ml
o-: Yeah, I could already hear your comment when I wrote this ;)

Isn't this the perfect case to get you used to overloading names. Due to the very obvious indentation structure I honestly don't think this would become more clear to read if I invented silly names for all the used steps... I can make you happy and make all of them anonymous :)

examples/inline_osr.sou
mhyee: Pick your favourite reason :)

- Many of our interesting examples have a before/after
- Run once and add it to repo, instead of running it each time I want to see it
- Maybe I want to show someone on my phone
- But really, I'm lazy

transform_assumption.ml
mhyee: Anonymous is much worse. :)

Can you rename the first one (line 60) to something like `used_checkpoints` (`used_targets`, etc)? Unlike the others, it's used all the way down on line 88, and it isn't obvious from the indentation.

transform_assumption.ml
mhyee: I'll live with this one.

transform_assumption.ml
o-: I have an alternative version that I like. But I am not sure if it is easier to understand now...

o-: The version now is in a sense nicer, but there is also beauty in stupid solutions

instr.ml
gasche: I am favorable to anti-over-engineering moves in general, but to be fair this is a case where the visitor has a *very* regular structure that always visits all arguments (I realize now that `op` is not visited, but that is a change you introduced), and I think moving away from this regular structure is wrong (before it was very simple to know what to do, and now people have to think about what to do in each case).

Besides, in that case visiting the label is precisely required if you want to freshen all the variables and labels of a piece of code. As for the `variable` case, we should have a generic `label` method that all three of `branch_label`, `goto_label` and `osr_label` defer to. I can send a PR after this one is merged.

o-: fair enough, though I must note that the original #op method did not have access to the arguments and I am really wondering how you would change the operation independently from the args :)

the not having to think and seeing the pieces in isolation was imho the reason for the bug where we converted `x[i]` to `nil[i]`.

I am not arguing against a visitor here. The only thing I am saying is that it is far from clear to me what the granularity of the visits should be and that if I am not sure how to proceed with a feature we don't need currently I leave it to be implemented when a usecase arises.

gasche: The issue was the mismatch between the type structure and what the parser allowed, not the visitor.

o-: ok, sorry I think I escalated this into a bigger argument than necessary. apologies. I'll try to restore the visitor as it was.

doc/inlining.md
```diff
@@ -0,0 +1,125 @@
...
+
+For the recreation of the stack to work the precise stacklayout of the caller needs to be known after the inlinee returns (ie. there must be an osr right after the call).
+The combined osr is supposed to push the continuation for the call to foo.
+
+```
+    2b  osr sp2 [] (foo, base, sp2) [var z = z_1],
+                   (main, base, sp1) [var y = $, var x = x]
+```
+
+The special % register specifies the target binding for the continuation.
```
mhyee: Typo: `%` -> `$`

examples/inline_osr.sou
o-: ok

transform.ml
o-: fixed (and there are people saying nobody reads commit msgs :) )

examples/inline_osr.sou
o-: (of course it revealed a bug)

o-:
ptal

parser.mly
```diff
@@ -145,8 +151,9 @@ instruction:
   label=label
   LBRACKET cond=separated_list(COMMA, expression) RBRACKET
   LPAREN func=label COMMA version=label COMMA pos=label RPAREN
-  LBRACKET map=separated_list(COMMA, osr_def) RBRACKET
-  { Osr {label; cond; target= {func; version; pos}; map} }
+  top=osr_map
+  frames=separated_list(COMMA, osr_frame)
+  { Osr {label; cond; target= {func; version; pos}; varmap=top; frame_maps=frames} }
```
gasche: I would rather like the first frame to also be separated from the osr_map by a comma. You can do this easily (and this can simplify `osr_frame` above as well) by just using `list(preceded(COMMA, foo))` instead of `separated_list(COMMA, foo)`. You can also give this pattern a name:

```
%inline preparated_list(sep, elem): list(preceded(sep, elem)) { $1 }
```

o-: thank you, I was looking for that trick!

o-:
@gasche are you happy with the rest?

gasche:
I haven't reviewed everything in detail, but I think the proposed design is solid and I am fine with merging the PR.
(I noticed a "it's => its" typo somewhere if you care.)

One merely stylistic comment: I think that multi-line `then/else` branches without `begin end` enclosing are dangerous (they work just fine as long as they start with a `let`, but they can surprise you if you start using sequences within their body), I think it would be good style to use `begin end` more often in these situations. No need to change the current patch, though.

Issue #159, opened 2017-06-06
Adding inline to the test runner


mhyee:
We added a new optimization pass (inlining), but didn't update the test runner.

I was going to submit a PR for this, but found another bug. I'm using this issue to store my notes so I can come back to this later.

These are the numbers on my computer. Looking at the most recent build (which does not test inlining), Travis takes almost 4 minutes to build everything and run `make test_examples_long`. (**Update**: Travis seems to take slightly over 4 minutes, with inlining.)

### Without inlining

`make test_examples`
```
real	0m3.975s
user	0m1.658s
sys	0m0.857s
```
`make test_examples_long`
```
real	0m7.474s
user	0m21.071s
sys	0m11.879s
```

### With inlining

`make test_examples`
```
real	0m6.080s
user	0m1.699s
sys	0m0.959s
```

`make test_examples_long`
```
real	0m20.611s
user	0m42.485s
sys	0m23.793s
```

mhyee:
Closed by #161.

Issue #158, opened 2017-06-06
Bug in param-call-01.sou with opts: prune,hoist_assign,inline


mhyee:
(I found this while adding `inline` to the test runner.)

```
$ ./sourir examples/param-call-01.sou --opt prune,hoist_assign,inline
After optimizations
function main ()
version inlined_version
 var x_1_1 = 0
 var res_1 = x_1_1
 # This test should print 0, 42, 0
 var x = 42
 var res = nil
 var x_1 = x
 assert (x_1_1 == 0)
 print x_1_1
 drop x_1
 var y = res_1
 drop res
 assert (x == 42)
 print x
 assert (y == 0)
 print y
 drop x
 drop x_1_1
 drop y
 drop res_1
 stop y
version anon
 [elided]
Scope error in the optimized program (prune, hoist_assign, inline):
Error in function main version inlined_version line 20: Variable y is not declared.
```

Note that `drop y` occurs before `stop y`.

I think this was introduced by #136...?

mhyee:
Closed by #161.

PR #161, opened 2017-06-06
Var bugfixes

o-:
the cleaner strikes again

### DIFF
--- a/test_examples.sh
+++ b/test_examples.sh
@@ -12,7 +12,7 @@ function ncores {
   fi
 }
 
-export OPTS="prune\nconst_fold\nhoist_assign\nhoist_drop\nmin_live"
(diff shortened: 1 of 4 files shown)


o-:
fixes #159 #158

mhyee (approved):
Thanks!

> But I noticed that inserting random variables into scope can only break the scope at merge points. There needs to be a goto right before the scope is broken. Inserting before goto is quite painless (since the goto itself does not interact with the scope). Thus the simple and conservative option is to only fix the scope when necessary right before a goto.

Yeah, I noticed something like this too, back when branch targets could be merge points. I had situations where I wanted to drop a variable `x` before the branch, so it would not be in scope after the branch. But then the branch condition required `x`...

PR #162, opened 2017-06-09
osr does not change the heap anymore

o-:
we should reflect this in the implementation since it simplifies
also the formal description of osr.

### DIFF
--- a/eval.ml
+++ b/eval.ml
@@ -264,15 +264,15 @@ let reduce conf =
     List.fold_left eval_arg Env.empty args
   in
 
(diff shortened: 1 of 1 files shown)


mhyee (approved):

gasche:
Is this a stable property or a temporary property? Wouldn't we want to be able to sink array allocations, and have `array x[0]` in the osr-map?

o-:
good question. I assume it's stable for this iteration since that would be a bigger change but maybe not...

PR #163, opened 2017-06-12
some features for the paper

o-:
1. Add a disassembler to produce ott compatible output that can be c/p into the tex
2. Untangle some optimizations to be able to show the changes in smaller increments.

### DIFF
--- a/ott.ml
+++ b/ott.ml
@@ -12,10 +12,10 @@ let rec dump_comma_separated how buf what =
   | e::t -> pr buf "%a, %a" how e (dump_comma_separated how) t
 
 let disassemble_instrs buf (instrs : instructions) =
-  let rec dump_instr buf pc needs_label =
-    if pc = Array.length instrs then ()
+  let rec dump_instr buf pc line needs_label =
+    if pc >= Array.length instrs then ()
     else begin
-      let dump_next_instr () = dump_instr buf (pc+1) true in
+      let dump_next_instr () = dump_instr buf (pc+1) (line+1) true in
       let simple buf = function
         | Var v             -> pr buf "%s" v
         | Constant c        -> pr buf "%s" (IO.string_of_value c)
@@ -42,15 +42,15 @@ let disassemble_instrs buf (instrs : instructions) =
         | Array_length e     -> pr buf "length(%a)" simple e
       in
       let dump_arg buf arg = dump_expr buf arg in
-      let print_pc_label () = pr buf "%d : " pc in
+      let print_default_label () = pr buf "%d : " line in
       let print_label label =
         let label = if String.length label = 1 then "l"^label else label in
         pr buf "%s : " label in
       begin match[@warning "-4"] instrs.(pc) with
-      | Label _ | Osr _ ->
+      | Label _ | Osr _ | Comment _ ->
         ()
       | _ ->
-        if needs_label then print_pc_label ()
+        if needs_label then print_default_label ()
       end;
       begin match instrs.(pc) with
       | Call (var, f, args)               ->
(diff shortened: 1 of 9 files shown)


transform.ml
```diff
@@ -130,22 +139,17 @@ let optimize (opts : string list) (prog : program) : program option =
...
-    if opts_assumpt <> []
-    then combine_opt [
+    combine_opt [
         (as_opt_program Transform_assumption.insert_checkpoints);
         optimizer;
-        (as_opt_program (as_opt_function Transform_assumption.hoist_assumption));
         optimizer;
         Transform_assumption.remove_empty_osr;
-        optimizer_classic;
+        optimizer;
```
mhyee: Remind me again, why do we need `optimizer` repeatedly here?

mhyee (commented):
It looks like you forgot to add some files (`ott.ml`?).

transform.ml
o-: no good answer...

some transformations might enable other, so running multiple times makes more optimizations.

maybe we could instead have a cli argument for it...

o-:
ok, now I added configurable optimizer pass number.
Also added the missing ott.ml file.
The code duplication is intentional since for now it would have been a bigger refactoring to make the disassembler configurable.

o-:
the ott syntax is a bit strange because I need to differ between normal instruction rendering and aligned instruction rendering.

The normal syntax of a program is:
```
fun1 -> (x) ( version1 -> ( l1 -> return x ) ) ),
fun2 -> (y) ( versionx -> ( l1 -> return y ) ) )
```

If you want aligned instructions (ie. the instruction rendered as `l1 & : & return x` you can enclose the program in `P( ... )`, or a plain list of labeled instructions in `i( ... )`.

ott.ml
```diff
@@ -0,0 +1,153 @@
...
+        pr buf " read %s\n" var;
+        dump_next_instr ()
+      | Osr {label; cond; target={func; version; pos}; varmap; frame_maps} ->
+        print_label label;
+        let dump_var buf = function
+          | Osr_var (x, e)     -> pr buf "%s = %a" x dump_expr e
+        in
+        let dump_frame buf {cont_pos={func; version; pos}; cont_res; varmap} =
+          pr buf "(%s, %s, %s) [%s = $, %a]"
+              func version pos
```
mhyee: Nitpick: indentation should be 2 spaces not 4. (See the `pr buf ...` below.)

ott.ml
```diff
@@ -0,0 +1,153 @@
...
+  in
+  dump_instr buf 0 true
+
+let disassemble buf (prog : Instr.program) =
+  let dump_func buf {name; formals; body} =
+      let print_formal buf (Param x) = pr buf "%s " x in
+      let print_formals buf = (dump_comma_separated print_formal) buf formals in
+      Printf.bprintf buf "%s -> ( %t ) ( " name print_formals;
+      let dump_vers buf version =
+          pr buf "%s -> (\n" version.label;
```
mhyee: Nitpick: should be 2 space indentation.

sourir.ml
```diff
@@ -13,6 +15,8 @@ let () =
     ("--quiet", Arg.Set quiet, "quiet");
     ("--autofix", Arg.Set autofix, "automatically normalize graph");
     ("--opt", Arg.String (fun s -> opts := String.split_on_char ',' s), "Enable optimizations");
+    ("--nopt", Arg.String (fun s -> opt_runs := int_of_string s), "Number of optimization runs");
```
mhyee: Could you rename this to `opt-runs` or `num-opts`? `nopt` looks too much like `nop` or `no opt` to me.

ott.ml
```diff
@@ -0,0 +1,153 @@
...
+      | Comment str                     ->
+        dump_next_instr ()
+      end;
+    end
+  in
+  dump_instr buf 0 true
+
+let disassemble buf (prog : Instr.program) =
+  let dump_func buf {name; formals; body} =
+      let print_formal buf (Param x) = pr buf "%s " x in
```
mhyee: Nitpick: should be 2 space indentation.

mhyee (approved):

PR #165, opened 2017-06-21
add annotations to inject assumptions

o-:
(this is mostly convenience to generate examples. review is not urgent, I wanted to publish it so you can all use it.)

guard_hint is a way of injecting an assumption.

in the assumption pass sourir will promote guard_hint into the nearest
condition for an osr instruction.

The instruction is called guard and not osr_hint because I would like to
rename osr to guard too in the future. More on that later...


analysis.ml
```diff
@@ -26,7 +26,7 @@ let successors_at (instrs : instructions) resolve pc : pc list =
   let instr = instrs.(pc) in
   let all_succ =
     match instr with
-    | Decl_var _ | Decl_array _
+    | GuardHint _ | Decl_var _ | Decl_array _
```
mhyee: `Guard_hint`

test_examples.sh
```diff
@@ -12,7 +12,7 @@ function ncores {
   fi
 }
 
-export OPTS="prune\nprune_no_hoist\nhoist_osr\nconst_fold\nhoist_assign\nhoist_drop\nmin_live\ninline_small\ninline_med"
+export OPTS="prune\nprune_no_hoist\nassumptions\nhoist_osr\nconst_fold\nhoist_assign\nhoist_drop\nmin_live\ninline_small\ninline_med"
```
mhyee, line 15: If we're adding this to the test runner, maybe we should have add a short program to the `examples` directory? The one in `tests.ml` should be enough.

tests.ml
```diff
@@ -1741,6 +1741,55 @@ let do_test_autofix () =
...
+    var x = 1
+    guard_hint (x==1)
+    branch (x==1) $la $lb
+    $la:
+      return 1
+    $lb:
+      return 2
+  |input} {expect|
+    function main ()
+    version anon_1
```
mhyee, line 1772: It looks like two versions are created. One which adds a checkpoint/guard/osr, and then the active version which removes the hint. In fact, it looks like a new version is created for each hint?

I think it'd be better if all the guards were added and hints removed in the same version. That way, there are only two versions: the new/active one, and the old/original one.

mhyee (commented):
Looked over this quickly. Merging/addressing comments is low priority.

tests.ml
o-: that's not sound. in the baseline the checkpoints need to be empty

analysis.ml
o-: ups, old habits

tests.ml
mhyee: I'm not sure I follow. In my proposal, the baseline is the original version, which has no checkpoints.

o-: the design decision was to insert the checkpoints in the baseline to be able to create new versions by just copying.

o-: all our passes do this

mhyee: OK, I see. Never mind.

test_examples.sh
o-: I'll do that in a later PR

Issue #40, opened 2017-01-31
Renaming "invalidate" into "guard".


gasche:
I gave a short presentation of speculative optimizations in general and some bits of Sourir IR to someone last week, and I found that using "invalidate" didn't really make sense to a newcomer, so I used "guard" instead. The problem with invalidate is that when reading

    invalidate (x > 0) foo

it is not clear if we invalidate if `(x > 0)` holds, or if it does not. On the other hard,

    guard (x > 0) foo

seems more clear, to me: we guard the fact that `(x > 0)` holds, so this is what the fast-path will assume. Of course, the status of the `foo` label is, conversely, a bit less clear (the "invalidation label" is clear, the "guard label" is not), but I think that this is overall less confusing. Note that we can have several keywords in our instruction, so `guard (x > 0) failwith l` for example would be possible (or `orelse`, `failto`, `or`). Another approach, relying on an extra keyword, would be `invalidate unless (x > 0)` (or `invalidate if (x <= 0)`), but having more negations makes things maybe a bit less clear.

Should we consider actually making the change?

o-:
I agree with invalidate being a bad name, especially with the condition being counter intuitive.

at least in the semantics doc we now changed it to "osr" and flipped the condition. Not sure if that was to your liking.

While I agree that for our current usecase "guard" would be the best I think in general "osr" is nicer. Because we can use it for osr-in and osr-out. Consider e.g.:

    osr (invocation_count > 100) optimized

Instead of the unless keyword I would propose to extend our primop with an unary negation. we would express "invalidate if x" or "guard not x" as "osr x" and "invalidate unless x" or "guard x" as "osr !(x)"

gasche:
I'm ok with `osr` and agree with the plan to use it in the implementation.

PR #166, opened 2017-06-27
Rename

o-:
Guard (like osr) is to narrow. it goes in the wrong direction. Our
approach is not "let's optimize, then add guards to safe us" but rather
"lets assumpe something then see what optimizations follow".

Also bailing out if any of the conditions is true (rather than if any of
them is false) seems counter intuitive for most people. So lets bite the
bullet and switch it.

My proposal is:

in sourir:
assumption l [guard,...] (t2..) [x=x,...], (t2...) [x...]

in ott:
l: assumption [guard,...] with (t2..) [x=x,...], (t2...) [x...]

This is the intended naming scheme:

* the whole thing is a checkpoint
* l is the label of the checkpoint
* [guard,...] are the guards or assumptions. the assumption is *positive*, we bail out if the guard *fails*  (new)
* (t2..) is the bailout target
* [x=x,...] is the variable map or environment map
* the rest are the extra frames, each extra frame again has a bailout target and a varmap

### DIFF
--- a/analysis.ml
+++ b/analysis.ml
@@ -29,7 +29,7 @@ let successors_at (instrs : instructions) resolve pc : pc list =
     | Decl_var _ | Decl_array _
     | Assign _ | Array_assign _
     | Drop _ | Read _ | Call _ | Label _
-    | Comment _ | Osr _ | Print _ | Assert _ ->
+    | Comment _ | Assumption _ | Print _ | Assert _ ->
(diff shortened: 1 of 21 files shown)


mhyee:
In our meeting, we decided to go with `assume (...) else (...)`.

(See also #40 for previous discussion.)

o-:
superseeded by #168

PR #167, opened 2017-06-28
Fix inlining

o-:


transform_prune.ml
```diff
@@ -2,7 +2,9 @@ open Instr
 open Types
 open Transform_utils
 
-let insert_branch_pruning_assumption ?(hoist=true) (func : afunction) : version option =
+let insert_branch_pruning_assumption ?(prune=true) ?(hoist=true) (func : afunction) : version option =
+  (* TODO support pruning false branch *)
+  assert (prune == true);
```
gasche: should be `prune = true`.

gasche (commented):
I don't really agree that the approach in 1c73c9754ab5ba is more robust. It is robust if we assume that the calls to *this* `fresh` function are the only ones that introduce new names, but it becomes fragile as soon as we mix several ways to introduce new names. I would rather have an explicit (and correct!) tracking of "what is currently in the environment" on the client side, which can be done simply by rematerializing the `instrs` after changes and passing that to further `fresh` calls.

That said, the current current is "more correct" than the previous implementation, and we need to move forward, so I would suggest to merge it even if I don't like the approach.

transform_assumption.ml
```diff
@@ -58,22 +58,33 @@ module TargetSet = Set.Make(LabelTarget)
...
 let remove_empty_osr : opt_prog = fun prog ->
-  let get_targets = function[@warning "-4"]
-    | Osr {target; cond; frame_maps} when cond <> [] ->
-      let add tgs {cont_pos} = TargetSet.union tgs (TargetSet.singleton cont_pos) in
-      List.fold_left add (TargetSet.singleton target) frame_maps
+  let get_targets func version seen = function[@warning "-4"]
+    | Osr {label; target; cond; frame_maps} ->
+      let my_pos = { func = func.name; version = version.label; pos = label } in
+      if cond <> [] || TargetSet.mem my_pos seen then begin
+        let add tgs {cont_pos} = TargetSet.union tgs (TargetSet.singleton cont_pos) in
```
gasche, line 65: This is orthogonal to the current change but I would like to be sure that I understand correctly. The reason that we also add the continuation targets to the live target set is not that we know they will be `osr`-ed to; they may not be, as we only know that a `return` instruction will want to jump to there, but because we will use the label->pc translation only at runtime to materialize the continuations when taking this osr, and thus we need to preserve the label to do this translation.

transform_assumption.ml
```diff
@@ -58,22 +58,33 @@ module TargetSet = Set.Make(LabelTarget)
...
-      extract (fun (v : version) ->
-        extract get_targets (Array.to_list v.instrs)) f.body) (prog.main::prog.functions)
+  (* Collect the targets (including the extra frames continuations) of all non empty osrs.
+   * This is a fixedpoint computation since empty osrs might be targets and thus need to be
+   * kept alive, even though they are empty. *)
+  let rec fixpoint_used seen =
+    let used =
+      extract (fun (f : afunction) ->
+        extract (fun (v : version) ->
+          extract (get_targets f v seen) (Array.to_list v.instrs)) f.body) (prog.main::prog.functions)
```
gasche, line 81: I'm not saying you should change, but here is the monadic-style version for reference:

```ocaml
let bind m f = extract f m in
bind (prog.main :: prog.functions) @@ fun (f : afunction) ->
bind f.body @@ fun (v : version) ->
bind (Array.to_list v.instrs) @@ fun instr ->
get_targets f v seen
```

gasche (commented):
>  also the extra frames can be dumped since we chain checkpoints
and the checkpoint one level down will be responsible for
creating them.

How come we didn't catch that with the testsuite? Maybe we should add some extra-frame tests?

gasche:
b74ff8f06baf8dd972dfb17985aad67818c42d98 is nice.

transform_assumption.ml
o-: sorry I don't understand the question.

The problem is that we could have the following:
non-empty-osr -> empty-osr -> empty-osr

o-:
I had to push another fix in 3f96c84. After that I hope it works

o-:
btw. i did add some tests in the examples

gasche:
Re. 887f83b81d55ccfa47c9bfed0dd734f28b086b96 : I don't understand enough of what is happening and I think it is time to go back to the drawing board (maybe for you to understand things better, but certainly for me at least). Can we go back to `doc/inlining.md`, work out an example where the assumptions being inlined already have stack frames, and discuss what the semantics of this and the correct transformation is?

Some questions that arose as I was looking at 887f83b81d55ccfa47c9bfed0dd734f28b086b96:

1. Are the stack frames environments to be understood in term of the inlined-assumption environment, or the after-the-first-jump environment? (I think: inlined-assumption)

2. What exactly does it mean for stack frames to use different versions, or a version different from the deopt target? (I think that this is fine, but I would like to see clear examples)

It may also be time to update `doc/ir-semantics` with the extra stack frames and the new assumption syntax.

gasche:
Note: right now it's a pain that there are so many interdependencies between the PRs. We want to merge the "renaming" PR, and the "prune moar" is low-risk, but they both depend on this PR that I think needs more time. Maybe we could just merge everything and do another iteration on the inlining semantics, possibly partially reverting this one as a result? Or could you / would you rebase the renaming / pruning PRs to not depend on this one?

gasche:
> The fix is to create new versions before inlining. That ensures that we always
prune the extra frames to 0 before inlining.

So if I understand correctly the fix you have in mind is "never inline assumptions with non-empty continuations"? That seems very strong, (1) we want to think about whether it would be correct to inline assumptions with non-empty continuations and (2) your proposed restriction would suggest that maybe the syntax should distinguish more things?

o-:
@gasche yeah I am sorry. My working hypothesis was that the inlining fix will be simple bugfix and that's why I put the renaming on top. I would rather not try and entangle it again. It's certainly less broken than before :(( so if we can merge that would probably be the better option.

but I agree that it's a mess and we need to understand it better

o-:
> So if I understand correctly the fix you have in mind is "never inline assumptions with non-empty continuations"? That seems very strong

yes, it's probably to strong. But currently I don't understand what invariant we are violating if we don't.
that's why thought that we should strengthen it first and then maybe loosen it again...

o-:
btw the repro was:

    ./sourir --opt prune_true,inline_small,inline_small,inline_small --num 1 rec.sou

with rec.sou:
```
function main ()
 #inline here
 call x = 'foo(10)
 print x
 return x

function foo(var x)
  var d = (x - 1)
  # this is inside foo
  branch (d != 0) $c $r
 $r:
  call r = 'foo(d)
  return r
 $c:
  return d
```

transform_assumption.ml
gasche: Right, but that only justifies adding the *assumption* target to the seen-set. My question was about the `cont_pos`, the labels that are not jumped to on deoptimization, but describe extra stack frames.

gasche:
I agree we can merge. I'm going to bike to Northeastern, get lunch etc. Feel free to merge in the meantime. But I think that we need to take a fresh look at the whole thing (the interaction between inlining and continuation frames) and I consider the current (post-PR) implementation broken (copying a new version is either a workaround or indicative of a design problem), and we need to understand this before we start (re)writing this part of the paper.

o-:
I think what breaks is the following:

before piling up the continuations we had the guarantee that the osr follows directly after the call, ie. the scope of the inlined callee is exactly the scope at the call instruction+ the scope of the original callee.

but then if you later on put other instruction between the call and the checkpoint this invariant is violated. so you have a continuation which relies on the fact that the checkpoint and call instruction have exactly the same scope (modulo the accumulator), but the actual code does something else.

o-:
on the other hand we have always had the invariant that you are not allowed to modify code with incomming deoptimization edges (that is quite reasonable, because otherwise you would need to update both ends in sync).

From that pov. it's natural that you cannot inline stuff into a function which is itself target of a deopt (that is exactly what happened...)

o-:
note, that you don't have to create new versions for every inlining. You can do multiple inlinings (otherwise the extra_frames would be max=1). But you cannot do inlining, then other changes, then inlining again on the modified version. Since inlining adds new guards it has to operate on a version where the checkpoints are up-to-date.

PR #171, opened 2017-06-28
don't rely on osr after call for extra_frames in inlinee

o-:
add an implicit bailout label to call instruction, use that
one for bailout targets of extra frames.


o-:
this is more or less how we described it in the paper so let's keep it for now.

PR #172, opened 2017-06-28
rewrite the inlining specification

gasche:
Today Oli and I discussed a relatively major misunderstanding we had
of the behavior of inlining of assumptions that contain extra
continuation frames. This new specification reflects my current
understanding of the situation -- that is, what the implementation
*should* do.

### DIFF
--- a/doc/inlining.md
+++ b/doc/inlining.md
@@ -55,6 +55,8 @@ function main()
     drop z'
     goto l
   # epilogue: return label
+  # the scope at `l`, because we dropped caller variables,
+  # is the same as in the non-inlined version
   l:
 
   print y
@@ -66,9 +68,10 @@ As a sidenote: If the call is not constant yet we could optimistically make it c
 
 ```
 function main()
+version base
   var x = 1
-  var call = someExpression
-  call y = call(x)
+  var fun = someExpression
+  lcall: call y = fun(x) -> lret
   print y
 ```
 
@@ -77,9 +80,9 @@ to:
 ```
 function main()
   var x = 1
-  var call = someExpression
-  assume [call == 'foo] or ...
-  call y = 'foo(x)
+  var fun = someExpression
+  assume [fun == 'foo] else (main, base, lcall) [var x = x, var fun = fun]
+  lcall: call y = 'foo(x) -> l
   print y
 ```
 
@@ -88,8 +91,8 @@ function main()
 Having the osr in the outer function poses no additional
 difficulties. Now lets assume the inlinee has safepoints.
 
-As an example (we use `rN` for calleR labels, `eN` for calleE label,
-and `iN` for Inlined-code labels):
+As an example (we use `rN` for "calle**r** labels", `eN` for
+"calle**e** label", and `iN` for "**i**nlined-code" labels):
 
 ```
 function main()
@@ -101,7 +104,7 @@ function main()
 
 function foo(var z)
   version base
-    e1:   assume [...] or (bar, base, checkpoint) [var z = z], cont-frames
+    e1:   assume [...] else (bar, base, checkpoint) [var z = z], cont-frames
     e2:   return z
 ```
 
@@ -114,7 +117,7 @@ function main()
         # prologue
         var z' = z
         var y = nil
-    i1: assume [...] or (bar, base, checkpoint) varmap, cont-frames # wrong
+    i1: assume [...] else (bar, base, checkpoint) varmap, cont-frames # wrong
     i2:
         y <- z'
         drop z'
@@ -123,11 +126,12 @@ function main()
         print y
 ```
 
-This is incorrect, because we would jump into (bar, base, checkpoint)
-with one less frame on the stack than previously. Before inlining, the
-bailout would end up with `main()`, then `foo()`, then the frames of
-`cont-frames` on the stack. After this broken inlining you would only
-get `main(), cont-frames`. We need to recreate a frame for `foo()`.
+This is incorrect, because on bailout we would jump into (bar,
+base, checkpoint) with one less frame on the stack than
+previously. Before inlining, the bailout would end up with `main()`,
+then `foo()`, then the frames of `cont-frames` on the stack. After
+this broken inlining you would only get `main(), cont-frames`. We need
+to recreate a frame for `foo()`.
 
 A continuation frame description contains:
 
@@ -135,8 +139,7 @@ A continuation frame description contains:
 
 - caller version (`main/base`)
 
-- the return label label `l`, which will be turned into a return pc
-  when building the frame at runtime
+- the return label `l`
 
 - the environment at the call point (or, really, the environment
   needed after returning)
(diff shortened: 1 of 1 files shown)


gasche:
cc @o- : let's make sure that we agree on this

(but you are also allowed to, you know, sleep)

o-:
I think we are in violent agreement.

I convinced myself that it works without the checkpoint and #171 is the implementation according to this spec.

o-:
I still have to read this PR in detail tomorrow...

doc/inlining.md
```diff
@@ -1,35 +1,61 @@
...
+  branch t $z $x
+  $z:
+    var z = x + x
+    print z
+    return z
+  $x:
+    return x
 ```
 
+(the `-> l` syntax indicates that `l` is the "return label" of the
```
o-: I would like to gloss over this in the spec. that seems like a strange complication. Since all of our instructions are labeled it seems we could cheat a bit by saying that in the case of `l: call ...` the deopt target is `l+1`, where the `+1` is done when it triggers.

doc/inlining.md
```diff
@@ -52,74 +78,122 @@ to:
 function main()
   var x = 1
   var call = someExpression
-  osr [call != 'foo] ...
+  assume [call == 'foo] or ...
```
o-: s/or/else

doc/inlining.md
```diff
@@ -52,74 +78,122 @@ to:
...
 
 ## With osr
 
-Having the osr in the outer function poses no additional difficulties. Now lets assume the inlinee has safepoints.
-As an example:
+Having the osr in the outer function poses no additional
+difficulties. Now lets assume the inlinee has safepoints.
+
+As an example (we use `rN` for calleR labels, `eN` for calleE label,
+and `iN` for Inlined-code labels):
```
o-: this confused me a bit since in `calleR` you highlight by capitalizing and in `rN` it's sort of the other way round. No changes needed, since in the paper it will be different presumably.

doc/inlining.md
```diff
@@ -52,74 +78,122 @@ to:
...
-
-function foo(var z)
-  version base
-    1   osr sp2 [] (foo, base, sp2) [var z = z]
-    2   return z
+        var x = 1
+        # prologue
+        var z' = z
+        var y = nil
+    i1: assume [...] or (bar, base, checkpoint) varmap, cont-frames # wrong
```
o-: here you switch syntax for the call (no ->). I like it more.

doc/inlining.md
```diff
@@ -52,74 +78,122 @@ to:
...
+        y <- z'
+        drop z'
+        goto l
+    l:
+        print y
 ```
 
-For the recreation of the stack to work the precise stacklayout of the caller needs to be known after the inlinee returns (ie. there must be an osr right after the call).
-The combined osr is supposed to push the continuation for the call to foo.
+This is incorrect, because we would jump into (bar, base, checkpoint)
```
o-: becaus /if the guard fails/ we would

doc/inlining.md
```diff
@@ -52,74 +78,122 @@ to:
...
-The special $ register specifies the target binding for the continuation.
+- the return variable `y`
 
-Remember the following defs from the semantic:
+- caller version (`main/base`)
 
-    C ::= (a, E, I, l)                 continuation:  result accumulator, environment, instructions, label
-    M ::= (P, C*) : (I, T, H, E, l)    machine state: program, continuations : instructions, trace, heap, environment, pc
+- the return label label `l`, which will be turned into a return pc
+  when building the frame at runtime
```
o-: in the formal version we don't have pc's. all labels are equal.

doc/inlining.md
```diff
@@ -52,74 +78,122 @@ to:
...
 
-If the osr fires the next state would be:
+- the environment of the caller, E{caller}
+- the environment of the inlined code, E{inlined}N
 
-    (P, (y, {x=1}, main::base, sp1)) :  (foo::base, T, H, {z=1},         1)
+At any label `iN`, the current environment is the disjoint union of
+`E{caller}` and `E{inlined}N`. We know that `E{caller}` is unchanged during
+the inlined function execution, given that those instructions come
+from the callee where it is not in scope, and thus cannot modify it.
```
o-: you could also argue that we make sure that at `l+1` (ie. after the inlined code) all inlinee vars are dropped and the scope is exactly that of the caller.

doc/inlining.md
```diff
@@ -52,74 +78,122 @@ to:
...
+```
+
+The bailout mapping `map` goes from the environment `E{callee}N` to
+the bailout-destination environment that we can call
+`E{checkpoint}`. In the inlined position, we need to use `ρ(varmap)`
+instead (applying `ρ` to every right-hand-side of `varmap`), that goes
+from `E{inlined}N` (or its extension, the environment of `eN`) into
+`E{checkpoint}` as expected.
+
+The continuation frames need to be updated in the same way.
```
o-: that's misleading, ρ should *only* be applied to the existing continuation frames and not to the *additional* ones, because they come from the outer scope.

doc/inlining.md
```diff
@@ -52,74 +78,122 @@ to:
...
+Let us consider again the bailout instruction to inline:
+
+```
+    e1:   assume [...] or (bar, base, checkpoint) varmap, cont-frames
+```
+
+The bailout mapping `map` goes from the environment `E{callee}N` to
+the bailout-destination environment that we can call
+`E{checkpoint}`. In the inlined position, we need to use `ρ(varmap)`
+instead (applying `ρ` to every right-hand-side of `varmap`), that goes
```
o-: since the osr map is part of the instruction stream and the rhs of the varmap are normal uses, a correct variable renaming implementation automatically does this. No need to stress it imho. You can say that applying ρ to the inlinee ρ(inlinee) will make sure that the varmap and con-frames are updated too.

doc/inlining.md
```diff
@@ -52,74 +78,122 @@ to:
...
+is statically known, it can be constant-folded after `l:`). But note
+that if we bail out of `foo`, it is precisely because some of its
+assumptions are invalid; we don't want to jump back into code that may
+be optimized based on assumptions that we now know are incorrect.
+
+To summarize:
+
+```
+    i1:   assume [ρ(...)] or (bar, base, checkpoint) ρ(varmap),
+            (y, main/base, l, E{caller}), ρ(cont-frames)
```
o-: perfect

o- (commented):
I like it

doc/inlining.md
gasche: I thought you liked the idea of calls having a return label in their structure. Why not use it in the spec as well? I personally feel that using `goto (l+1)` for each `return` in the callee body is sort of inelegant. (If there was just one occurrence, fine, but we use it several time, I would rather have a proper name for the label.) If we cannot have `-> l`, I'll just consistently label the next expression `l` in all my examples.

doc/inlining.md
gasche: Good point, I'll use "calle**r**" instead of "calleR".

doc/inlining.md
gasche: Wait, this specific example had the call inlined, so there is no call instruction anymore and thus no `-> l`. One example above (in the function-specialization case, where no inlining is done) didn't have the return label, but I added it for your sake :-)

doc/inlining.md
gasche: I think this is orthogonal: this section is talking carefully about how the scopes evolve on `osr` instructions inlined. Your remark concerns the fact that return instructions (explicit or implicit) are translated in a scope-preserving way, but it is more about the correctness of inlining in general than osr-specific.

I added a remark about that in the comments of the warmup example.

doc/inlining.md
gasche: I added a remark to this effect at the end of the document.

doc/inlining.md
gasche: I reformulated, thanks.

doc/inlining.md
```diff
@@ -51,75 +80,128 @@ to:
...
 
 ## With osr
 
-Having the osr in the outer function poses no additional difficulties. Now lets assume the inlinee has safepoints.
-As an example:
+Having the osr in the outer function poses no additional
+difficulties. Now lets assume the inlinee has safepoints.
+
+As an example (we use `rN` for "calle**r** labels", `eN` for
+"calle**e** label", and `iN` for "**i**nlined-code" labels):
```
gasche, line 95: Markdown fails to render this correctly in diff mode (I hoped that the quote I added would fix it), but it renders correctly otherwise as a markdown document, so let's go for that.

doc/inlining.md
o-: I like it for the implementation because it ensures that you cannot accidentially insert instructions between the call and the label.
But for the formalism I find it not necessary and honestly to me it stands out (ie. why do you even need it, there is a label guaranteed to be right after the call...)

o-:
lgtm

except for the goto label discussion. It's you call. If you think you can do the proofs by using the normal labels (or by inserting an extra normal label after the call) I would go this way, since it's one thing less.

If you think it simplifies the argument a lot. Then I would keep it.

gasche:
I like having return labels be part of the syntax, but I hid them for now.

PR #173, opened 2017-06-28
refactoring in the inliner regarding extra_frames

o-:
the current inliner stores all intermediate versions of all inlinees.
that is unnecessary conservative.

in fact if our inlining approach to always restore the full call
stack (without jumping back to caller with inlinings) works
then it must not be necessary to keep them.


o-:
this is a minor fix to the current inliner, but I think we should actually merge #171 instead

PR #174, opened 2017-07-03
osr-in example

gasche:
From a discussion of Oli who suggested a trick from Daniele (who is Daniele?)

cc @o-


mhyee:
Probably http://www.dis.uniroma1.it/~delia/

o-:
yes, http://www.dis.uniroma1.it/~delia/papers/cgo2016-osrkit.pdf

o-:
it's a bit different in the paper, but let's keep it for now for reference

